PR #40, opened 2019-09-21
Scrub messages

aygold92:
*Issue #, if available:*

*Description of changes:*
Ensure that during schema validation, no messages contain values.

The strategy is to use keywords and change the error message for the ones that emit values.

Certain keywords were whitelisted as safe, in case we start using a new keyword. All other keywords fail with message:
`<SchemaPointer>: failed validation constraint for keyword [<keyword>]`

The following emit values:
* Integer Keywords: `multipleOf`, `minimum`, `maximum`, `exclusiveMaximum`, `exclusiveMinimum`
* String Keywords: `pattern`
* `enum`
* `const`

Bumped the version because it has backwards incompatible changes (changed schemaLocation -> schemaPointer).  Will need to fix the java plugin as well

### DIFF
--- a/src/test/java/com/amazonaws/cloudformation/resource/ValidatorTest.java
+++ b/src/test/java/com/amazonaws/cloudformation/resource/ValidatorTest.java
@@ -310,7 +310,7 @@ public class ValidatorTest {
             .put(PRIMARY_IDENTIFIER_KEY, Arrays.asList(EXAMPLE_PRIMARY_IDENTIFIER));
 
         assertThatExceptionOfType(ValidationException.class).isThrownBy(() -> validator.validateResourceDefinition(definition))
-                .withNoCause().withMessage("#/properties: minimum size: [1], found: [0]");
+            .withNoCause().withMessage("#/properties: minimum size: [1], found: [0]");
     }
 
     @Test


PatMyron:
This currently scrubs too much information. In addition to scrubbing the property value, it also removes the constraint itself, which is the most important context

aygold92:
> This currently scrubs too much information. In addition to scrubbing the property value, it also removes the constraint itself, which is the most important context. Is it possible to only scrub [`writeOnlyProperties`](https://github.com/aws-cloudformation/cloudformation-resource-schema#resource-semantics)?

What do you mean it removes the constraint? Is there a bug? As noted in description, the message *should* be `<SchemaPointer>: failed validation constraint for keyword [<keyword>]`

PatMyron:
What is the actual `pattern`/`enum`/`minimum`/`maximum`/etc.? That's the most important context. Knowing which `pattern` you failed is far more useful than just knowing you failed some `pattern` that's been scrubbed

aygold92:
> What is the actual `pattern`/`enum`/`minimum`/`maximum`/etc.? That's the most important context. Knowing which `pattern` you failed is far more useful than just knowing you failed some `pattern` that's been scrubbed, and we could reveal these without revealing property values:
> 
> ```
> **** is greater than the maximum of 100
> **** is less than the minimum of 0
> '****' is not one of ['a', 'b', 'c']
> '****' does not match 'arn:.*'
> ```

True, this could be a better error message, I would support improving it.  That being said, for now they can always check the schema by calling DescribeType

PR #47, opened 2019-10-22
default value for insertionOrder is true

aygold92:
*Issue #, if available:*

*Description of changes:*
The default value for insertionOrder should be true, because by default we will honor order (i..e, we assume list by default, rather than set). This kind of makes me think the property should actually be called something else, so that the default value can be false, but I'm not sure we can make that change at this point. If so, would like to have suggestions (my original idea was "orderMatters", which allows us to default to false, but that sounds bad)

### DIFF
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -76,7 +76,7 @@
                         "insertionOrder": {
                             "description": "When set to true, this flag indicates that the order of insertion of the array will be honored, and that changing the order of the array would indicate a diff",
                             "type": "boolean",
-                            "default": false
+                            "default": true
                         },
                         "$ref": {
                             "$ref": "https://json-schema.org/draft-07/schema#/properties/$ref"


PatMyron:
Description reasoning feels somewhat circular to me, don't completely follow why the `default` definition switched to `true` rather than the implementation's `default` switching to `false` to match the previous definition

Less than 200 of the 1000+ [public resource schema](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) [`arrays`](https://json-schema.org/understanding-json-schema/reference/array.html) specified [`insertionOrder`](https://github.com/aws-cloudformation/cloudformation-resource-schema#insertionorder), and [`insertionOrder`](https://github.com/aws-cloudformation/cloudformation-resource-schema#insertionorder) defaulting to `true` is one of the main issues with [drift detection for Registry types](https://aws.amazon.com/about-aws/whats-new/2020/10/aws-cloudformation-drift-detection-supports-cloudformation-registry-resource-types/) causing false positives

Seems easier to `default` to `false` than modifying most resource schemas to specify [`insertionOrder`](https://github.com/aws-cloudformation/cloudformation-resource-schema#insertionorder) for drift detection to work properly

aygold92:
> Description reasoning feels somewhat circular to me, don't completely follow why the `default` definition switched to `true` rather than the implementation's `default` switching to `false` to match the previous definition

> Seems easier to `default` to `false` than modifying most resource schemas to specify [`insertionOrder`](https://github.com/aws-cloudformation/cloudformation-resource-schema#insertionorder) for drift detection to work properly

Not sure what you mean by this. By default, order is honored, and if they have a Set then they need to say `insertionOrder: false`.  It would have been better to call this something where the default could be false, but in terms of updating schemas that have not included it, it doesn't really make a difference.  In most cases, teams don't need to include it because by default we honor the order

PR #60, opened 2019-11-22
Allow use of $id keyword in resource schema

vladtsir:
*Issue #, if available:*

This change allows to use $id keyword in resource schema, which is necessary to support relative refs to other schemas, which in turn is needed for schema regionalization.

### DIFF
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -212,9 +212,9 @@
     "type": "object",
     "patternProperties": {
         "^\\$id$": {
-            "$ref" : "https://json-schema.org/draft-07/schema#/properties/$id"
+            "$ref": "https://json-schema.org/draft-07/schema#/properties/$id"
         }
-    },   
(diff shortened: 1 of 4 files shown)


rjlohan:
Agree on adding test and fixing build.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -210,6 +210,11 @@
         }
     },
     "type": "object",
+    "patternProperties": {
+        "^\\$id$": {
+            "$ref": "https://json-schema.org/draft-07/schema#/properties/$id"
+        }
+    },
```
aygold92: why is this expressed via patternProperties? Couldn't there just be a `"$id"` key under properties?

tobywf: we tried that and it doesn't work, because `$id` is a valid field that can appear anywhere: https://github.com/aws-cloudformation/cloudformation-cli/pull/344#issuecomment-554839461

PR #61, opened 2019-11-26
add custom stabilization timeout in handlers schema

aygold92:
*Issue #, if available:*

*Description of changes:*
This adds a new parameter `timeoutInMinutes` to the `handlerDefinition`, in order for providers to specify a custom stabilization timeout, to be used by CloudFormation. This could be used by the handler itself at some point to stop doing work.

I used `timeoutInMinutes` since it is much simpler for handlers rather than specifying some ISO format, and I don't see any need to have any more granularity than that

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ For each handler, you should define a list of API `permissions` required to perf
 
 ### timeoutInMinutes
 
-For each handler, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.
+For create, update, and delete handlers, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ For each handler, you should define a list of API `permissions` required to perf
 
 ### timeoutInMinutes
 
-For create, update, and delete handlers, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.
+For each handler, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.  Currently, this value is only used for `Create`, `Update`, and `Delete` handlers, while `Read` and `List` handlers are expected to return synchronously within 30 seconds.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ We have taken an opinion on certain aspects of the core JSON Schema and introduc
 ### New Schema-Level Properties
 
 #### insertionOrder
+
 Array types can define a boolean `insertionOrder`, which specifies whether the order in which elements are specified should be honored when processing a diff between two sets of properties.  If `insertionOrder` is true, then a change in order of the elements will constitute a diff.  The default for `insertionOrder` is true.
 
 Together with the `uniqueItems` property (which is native to JSON Schema), complex array types can be defined, as in the following table:
@@ -138,6 +139,18 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 * **`additionalProperties`** use of `additionalProperties` is not valid for a resource property. Use `patternProperties` instead to define the shape and allowed values of extraneous keys.
 * **`properties` and `patternProperties`** it is not valid to use both properties and patternProperties together in the same shape, as a shape should not contain both defined and undefined values. In order to implement this, the set of undefined values should itself be a subshape.
 * **`items` and `additionalItems`** the `items` in an array may only have one schema and may not use a list of schemas, as an ordered tuple of different objects is confusing for both developers and customers. This should be expressed as key:value object pairs. Similarly, `additionalItems` is not allowed.
+
+## handlers
+
+The `handlers` section of the schema allows you to specify which CRUDL operations (create, read, update, delete, list) are available for your resource, as well as some additional metadata about each handler.
+
+### permissions
+
+For each handler, you should define a list of API `permissions` required to perform the operation.  Currently, this is used to generate IAM policy templates and is assumed to be AWS API permissions, but you may list 3rd party APIs here as well.
+
+### timeoutInMinutes
+
+For each handler, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.
 
 ## License
 
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -10,7 +10,7 @@
             "maxLength": 4096
         },
         "cudHandlerDefinition": {
-            "description": "Defines any execution operations which can be performed on this resource provider",
+            "description": "Defines create, update, and delete operations which can be performed on this resource provider",
             "type": "object",
             "properties": {
                 "permissions": {
@@ -30,7 +30,7 @@
             ]
         },
         "rlHandlerDefinition": {
-            "description": "Defines any execution operations which can be performed on this resource provider",
+            "description": "Defines read and list operations which can be performed on this resource provider",
             "type": "object",
             "properties": {
                 "permissions": {
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -19,6 +19,13 @@
                         "type": "string"
                     },
                     "additionalItems": false
+                },
+                "timeoutInMinutes": {
+                    "description": "Defines the timeout for the entire operation to be interpreted by the invoker of the handler.  The default is 120 (2 hours).",
+                    "type": "integer",
+                    "minimum": 2,
+                    "maximum": 4320,
+                    "default": 120
                 }
             },
             "additionalProperties": false,
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -9,22 +9,18 @@
             "pattern": "^https://[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
             "maxLength": 4096
         },
-        "handlerDefinition": {
+        "cudHandlerDefinition": {
             "description": "Defines any execution operations which can be performed on this resource provider",
             "type": "object",
             "properties": {
                 "permissions": {
-                    "type": "array",
-                    "items": {
-                        "type": "string"
-                    },
-                    "additionalItems": false
+                    "$ref": "#/definitions/permissionsDefinition"
                 },
                 "timeoutInMinutes": {
                     "description": "Defines the timeout for the entire operation to be interpreted by the invoker of the handler.  The default is 120 (2 hours).",
                     "type": "integer",
                     "minimum": 2,
-                    "maximum": 4320,
+                    "maximum": 720,
                     "default": 120
                 }
             },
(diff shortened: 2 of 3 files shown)


rjlohan:
Until the platform is capable of actually respecting this information I don't think it should be part of the schema.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -19,6 +19,13 @@
                         "type": "string"
                     },
                     "additionalItems": false
+                },
+                "timeoutInMinutes": {
+                    "description": "Defines the timeout for the entire operation to be interpreted by the invoker of the handler.  The default is 120 (2 hours).",
+                    "type": "integer",
+                    "minimum": 2,
+                    "maximum": 4320,
```
rjlohan: We can't support 72hours with Role based auth. 12h is the maximum supported by IAM/STS.

tobywf: think we also need to be clear which operations are subject to this timeout (CUD, not RL), right?

aygold92: that's true it would not apply to read and list... I'll have to split it into 2 handler definitions then

aygold92: > We can't support 72hours with Role based auth. 12h is the maximum supported by IAM/STS.

we have to support 72 hours at some point to maintain parity with resources we currently support, but I can lower it to 12 hours for the time being

aygold92:
> Until the platform is capable of actually respecting this information I don't think it should be part of the schema.

wanted to get the idea out there, but I will wait to merge this until its actually supported in the CFN platform

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -9,23 +9,46 @@
...
-                    },
-                    "additionalItems": false
+                    "$ref": "#/definitions/permissionsDefinition"
+                },
+                "timeoutInMinutes": {
+                    "description": "Defines the timeout for the entire operation to be interpreted by the invoker of the handler.  The default is 120 (2 hours).",
+                    "type": "integer",
+                    "minimum": 2,
+                    "maximum": 720,
+                    "default": 120
```
WaelA: Isn't this for IAM roles and these should range from [1 hour to 12 hours](https://aws.amazon.com/blogs/security/enable-federated-api-access-to-your-aws-resources-for-up-to-12-hours-using-iam-roles/)?

tobywf: 720s = 12h? don't see any indication in that link that less than 1h could be a problem?

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -323,6 +323,61 @@ public void validateDefinition_invalidHandlerSection_shouldThrow() {
...
+            .getResourceAsStream("/valid-with-handlers.json")));
+
+        final JSONObject createDefinition = definition.getJSONObject("handlers").getJSONObject("create");
+        createDefinition.put("timeoutInMinutes", timeout);
+
+        final String keyword = timeout == 1 ? "minimum" : "maximum";
+
+        assertThatExceptionOfType(ValidationException.class).isThrownBy(() -> validator.validateResourceDefinition(definition))
+            .withNoCause().withMessage(
+                String.format("#/handlers/create/timeoutInMinutes: failed validation constraint for keyword [%s]", keyword));
```
tobywf: `.withMessageContaining("#/handlers/create/timeoutInMinutes")` is fine. the reason we check the message at all is to make it easy to debug in future if the test fails for any other schema error. if you really want, `.withMessageContaining("#/handlers/create/timeoutInMinutes").withMessageContaining(keyword)` is extra good. but `.withMessage` is too brittle, requires too much maintenance.

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -323,6 +323,61 @@ public void validateDefinition_invalidHandlerSection_shouldThrow() {
...
+    public void validateDefinition_noTimeoutAllowed_shouldThrow(final String handlerType) {
+        final JSONObject definition = new JSONObject(new JSONTokener(this.getClass()
+            .getResourceAsStream("/valid-with-handlers.json")));
+
+        final JSONObject handlerDefinition = definition.getJSONObject("handlers").getJSONObject(handlerType);
+        handlerDefinition.put("timeoutInMinutes", 30);
+
+        assertThatExceptionOfType(ValidationException.class).isThrownBy(() -> validator.validateResourceDefinition(definition))
+            .withNoCause()
+            .withMessage(String.format("#/handlers/%s: extraneous key [timeoutInMinutes] is not permitted", handlerType));
```
tobywf: same as above - `.withMessageContaining("timeoutInMinutes")` should be fine

src/main/resources/schema/provider.definition.schema.v1.json
aygold92: this is not just for IAM roles, this is for the workflow timeout.  Even if the minimum role was an hour the handler could still timeout after 2 minutes.

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
aygold92: will do

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
aygold92: will do

benkehoe:
This shouldn't necessarily be static, right? With the [DynamoDB GSI feature request](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/229), the length of time is highly dependent on the data and usage. So instead you might want the resource to be able to report an expected completion time dynamically, which could then also be reported to the user.

ikben:
I think a static (maybe not required?) "maximum" could be useful, as a dynamic value would need an extra data to be returned by the resource (increasing complexity for handler that are fine with a static value, eg. code that never returns InProgress).

For dynamic values it might be interesting to have both a "expected completion" and a "pause time" value that can be returned. 

The expected completion could be exposed to the end user (a "completion percentage" might be another option for this).
The pause time could indicate that CloudFormation should not invoke the provider for a certain amount of time. While CloudFormation waits the cost  for going over the 30s threshold could be lower (ideally 0)

aygold92:
right, this is intended to signify the maximum timeout for a resource. A separate protocol could be developed using progress events to improve stabilization and possibly add dynamic timeouts (though dynamic timeouts could only shorten the timeout, and would still be subject to this maximum timeout)

tobywf:
the maximum is especially useful for the failure case. e.g. if you're developing a resource, right now if you get delete wrong you have to wait 2h until it times out. for normal operation, it doesn't affect the provisioning duration.

you can already "sleep" the handler via the callback. true, it would be nice to have a dynamic system where the first invocation has a timeout of a few minutes, and then can request/guesstimate further timeouts as required via the callback. but this is tricky w.r.t. credentials. i think the approach Andrew's proposing is a decent compromise for now?

PatMyron:
> This shouldn't necessarily be static, right? With the [DynamoDB GSI feature request](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/229), the length of time is highly dependent on the data and usage. So instead you might want the resource to be able to report an expected completion time dynamically, which could then also be reported to the user.

The completion time also depends on some factors that cannot be known ahead of time. Both of these factors can change drastically at any point during DynamoDB GSI creation:

> * The provisioned write capacity of the index
> * Write activity on the main table during index builds

benkehoe:
> Both of these factors can change drastically at any point during DynamoDB GSI creation

This shows that not only must it be dynamic, but that it also must be modifiable during the operation.

rjlohan:
This PR is only implementing a _default maximum_ time which CloudFormation should respect when terminating non-responsive, or incomplete handlers. This timeout is treated as a `FAILED` operation. Hence, there is already dynamism built into the platform for any handler - if a handler decides that it has run longer than "it should" (e.g; "well this GSI creation is taking longer than I expected...") it can return a `FAILED` response at any time during operation with some sort of `TimeoutExceeded` exception or message. 

I don't see any value in trying to introduce dynamic timeouts to either the schema or the interface, over the capabilities we already have.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -9,23 +9,46 @@
             "pattern": "^https://[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
             "maxLength": 4096
         },
-        "handlerDefinition": {
+        "cudHandlerDefinition": {
```
jotompki: i feel like asyncHandlerDefinition vs syncHandlerDefinition might be a better name. Also should probably update the description on the object in either case so that they are distinguishable

README.md
```diff
@@ -139,6 +140,18 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+
+The `handlers` section of the schema allows you to specify which CRUDL operations (create, read, update, delete, list) are available for your resource, as well as some additional metadata about each handler.
+
+### permissions
+
+For each handler, you should define a list of API `permissions` required to perform the operation.  Currently, this is used to generate IAM policy templates and is assumed to be AWS API permissions, but you may list 3rd party APIs here as well.
+
+### timeoutInMinutes
+
+For each handler, you may define a `timeoutInMinutes` property, which defines the *maximum* timeout of the operation.  This timeout is used by the invoker of the handler (such as CloudFormation) to stop listening and cancel the operation.  Note that the handler may of course decide to timeout and return a failure prior to this max timeout period.
```
jotompki: This clashes with our schema definition. If they add a timeout to read handler, schema won't be valid. Better to be clearer here

aygold92: ah yeah good catch, will specify for CUD actions only

src/main/resources/schema/provider.definition.schema.v1.json
aygold92: I'll update the description, not sure if async vs sync might be confusing though because a CUD handler could be sync or async, depending on how quickly it finishes

benkehoe:
So this timeout isn't a timeout for the operation, but for a specific invocation of the handler? That is, I set the timeout to 30 seconds, it's fine for the operation to take five minutes as long as my handler keeps returning a progress report in less than 30 seconds when it gets invoked (until the operation is complete)?

aygold92:
> So this timeout isn't a timeout for the operation, but for a specific invocation of the handler? That is, I set the timeout to 30 seconds, it's fine for the operation to take five minutes as long as my handler keeps returning a progress report in less than 30 seconds when it gets invoked (until the operation is complete)?

No, this timeout is the max possible time the entire operation would take.  So let's say you set the timeout to 5 minutes: 
CFN will invoke the handler and then listen to progress events for 5 minutes.  If the operation is still IN_PROGRESS after 5 mnutes, CFN will stop listening and will terminate the operation.

benkehoe:
> this timeout is the max possible time the entire operation would take.

So then I don't understand @rjlohan's comment that the purpose is "terminating non-responsive, or incomplete handlers"

Does a successfully-returning handler reporting IN_PROGRESS count as non-responsive? If I don't explicitly set the timeout value in the schema, presumably CloudFormation has some default setting; is there a way to know programmatically that the timeout for an operation is the default and what the default is?

rjlohan:
> Does a successfully-returning handler reporting IN_PROGRESS count as non-responsive? If I don't explicitly set the timeout value in the schema, presumably CloudFormation has some default setting; is there a way to know programmatically that the timeout for an operation is the default and what the default is?

No, that'd fall into the responsive, but incomplete handler bucket. e.g; a type may periodically check that it has stabilized, but perpetually fail to do so. Then the timeout gives CloudFormation a signal that it should stop waiting.

src/main/resources/schema/provider.definition.schema.v1.json
rjlohan: Need to leave this as it was. `handlerDefinition` is meant to be an abstract action concept. `Read` and `List` operations don't _need_ to define a timeout, but they could and it would still be valid. The `timeoutInMinutes` property should be optional anyway.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -282,19 +305,19 @@
             "type": "object",
             "properties": {
                 "create": {
-                    "$ref": "#/definitions/handlerDefinition"
+                    "$ref": "#/definitions/cudHandlerDefinition"
```
rjlohan: See above, we don't want to start schematising specialised handler operations. That's not a good abstraction.

benkehoe:
>  responsive, but incomplete handler

I get it now. Thanks! I still am wondering how I can know programmatically what the default timeout is.

src/main/resources/schema/provider.definition.schema.v1.json
aygold92: my idea here was that `Read` and `List` should not be able to define a timeout, and that CUD operations were meant to be async (though as we know could possibly finish synchronously), where as read and list were meant to be synchronous only.  In which case, a timeout would not make sense.  I think its a very bad experience to have an async describe call.  Keep in mind also that this is the two way door decision and could always be changed back later, whereas if we start letting customers define Read and List timeouts (which aren't used in any way) then we have to support that forever

rjlohan: Async or Sync response is an orthogonal concern to timeouts. Both can have a maximum wait time. Nothing in the schema indicates the calling mode for CRUDL operations. What I don't want is to start changing the schema to specialise the way handler actions are defined. These are _designed_ in the schema to be descriptions of arbitrary actions that can be taken on a type. They have a name, but they are just actions. It's up to the caller to understand the interface and the contract, including whether to call and wait synchronously, what the action payloads are, etc. The schema shouldn't attempt to start trying to define that interface and specialise each operation's definition - that really is a one-way door. 

On top of that, changing this now invalidates every existing schema that has already been published. This is not the right change. There is absolutely nothing wrong with having optional timeouts on all handlers which may or may not be honored by the caller for certain operations. You are defining a _maximum_ timeout, we can wait less as we choose.

PR #64, opened 2019-12-12
Adding methods to remove writeOnlyProperties from model

jotompki:
*Issue #, if available:*

*Description of changes:* Adding some methods to the ResourceSchema class to determine if a given model has writeOnlyProperties and also a method to remove the properties. Tried to make it extensible so that the same thing could be applied to other groups of properties.

I did not handle cases for writeOnlyProperties that were items in arrays, as I was unsure on if this was allowed or what the behavior should be. Say, if "/properties/arrayProperty/0" is writeOnly, and the array has two items, if we remove one, then the object still appears to have a writeOnlyProperty after removal:
```
{
   "arrayProperty": [0, 1]
}
```
After:
```
{
   "arrayProperty": [1] //still has item at "/properties/arrayProperty/0"
}
```
Input on this would be great.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -15,6 +15,7 @@
 package software.amazon.cloudformation.resource;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -23,7 +24,9 @@ import java.util.stream.Collectors;
 
 import lombok.Getter;
 
+import org.apache.commons.lang3.RegExUtils;
 import org.everit.json.schema.JSONPointer;
+import org.everit.json.schema.JSONPointerException;
 import org.everit.json.schema.ObjectSchema;
 import org.everit.json.schema.loader.SchemaLoader;
 import org.json.JSONObject;
@@ -144,5 +147,49 @@ public class ResourceTypeSchema extends ObjectSchema {
     @Override
     public Map<String, Object> getUnprocessedProperties() {
         return Collections.unmodifiableMap(this.unprocessedProperties);
+    }
+
+    public boolean hasWriteOnlyProperties(final JSONObject resourceModel) {
+        return this.getWriteOnlyPropertiesAsStrings().stream()
+            .anyMatch(writeOnlyProperty -> hasProperty(RegExUtils.removeFirst(writeOnlyProperty, "/properties"), resourceModel));
+    }
+
+    public boolean hasProperty(final String property, final JSONObject resourceModel) {
+        return hasProperty(new JSONPointer(property), resourceModel);
+    }
+
+    public boolean hasProperty(final JSONPointer property, final JSONObject resourceModel) {
+        try {
+            return property.queryFrom(resourceModel) != null;
+        } catch (JSONPointerException e) {
+            return false;
+        }
+    }
+
+    public void removeWriteOnlyProperties(final JSONObject resourceModel) {
+        this.getWriteOnlyPropertiesAsStrings().stream()
+            .forEach(writeOnlyProp -> removeProperty(new JSONPointer(RegExUtils.removeFirst(writeOnlyProp, "/properties")),
+                resourceModel));
+    }
+
+    public static void removeProperty(final JSONPointer property, final JSONObject resourceModel) {
+        String refParts = property.toString();
+        // split pointer into tokens, removing first delimiter
+        List<String> refTokens = Arrays.asList(RegExUtils.removeFirst(refParts, "/").split("/"));
+        final String key = refTokens.get(refTokens.size() - 1);
+        try {
+            // if size is more than one, fetch parent object/array of key to remove so that
+            // we can remove
+            if (refTokens.size() > 1) {
+                // use sublist to specify to point at the parent object
+                final JSONPointer parentObjectPointer = new JSONPointer(refTokens.subList(0, refTokens.size() - 1));
+                final JSONObject parentObject = (JSONObject) parentObjectPointer.queryFrom(resourceModel);
+                parentObject.remove(key);
+            } else {
+                resourceModel.remove(key);
+            }
+        } catch (JSONPointerException | NumberFormatException e) {
+            // do nothing, as this indicates the model does not have a value for the pointer
+        }
     }
 }
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -149,9 +149,8 @@ public class ResourceTypeSchema extends ObjectSchema {
     }
 
     public void removeWriteOnlyProperties(final JSONObject resourceModel) {
-        this.getWriteOnlyPropertiesAsStrings().stream()
-            .forEach(writeOnlyProperty -> removeProperty(new PublicJSONPointer(writeOnlyProperty.replaceFirst("/properties", "")),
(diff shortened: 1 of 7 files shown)


rjlohan:
Agree we don't need to support array items here. Either the whole array property is writeOnly or not.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -145,4 +148,48 @@ public static ResourceTypeSchema load(final JSONObject schemaJson) {
...
+                resourceModel));
+    }
+
+    public static void removeProperty(final JSONPointer property, final JSONObject resourceModel) {
+        String refParts = property.toString();
+        // split pointer into tokens, removing first delimiter
+        List<String> refTokens = Arrays.asList(RegExUtils.removeFirst(refParts, "/").split("/"));
+        final String key = refTokens.get(refTokens.size() - 1);
+        try {
+            // if size is more than one, fetch parent object/array of key to remove so that
```
rjlohan: Thinking maybe that rather than remove the property and give the appearance that it's not set at all, we redact/mask the value instead - this would indicate clearly that the property is set but not what its value is. Thoughts?

jotompki: Yeah that makes sense, but might get tricky if the fields have patterns/min/max length constraints, though I am not sure if that would be a common use case. The model after being used could potentially not be valid against the schema, though the same thing may be true from removing a property.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -145,4 +148,48 @@ public static ResourceTypeSchema load(final JSONObject schemaJson) {
     public Map<String, Object> getUnprocessedProperties() {
         return Collections.unmodifiableMap(this.unprocessedProperties);
     }
+
+    public boolean hasWriteOnlyProperties(final JSONObject resourceModel) {
```
tobywf: what's this function for? can't we just strip them if they exist?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -145,4 +148,48 @@ public static ResourceTypeSchema load(final JSONObject schemaJson) {
...
+        try {
+            return property.queryFrom(resourceModel) != null;
+        } catch (JSONPointerException e) {
+            return false;
+        }
+    }
+
+    public void removeWriteOnlyProperties(final JSONObject resourceModel) {
+        this.getWriteOnlyPropertiesAsStrings().stream()
+            .forEach(writeOnlyProp -> removeProperty(new JSONPointer(RegExUtils.removeFirst(writeOnlyProp, "/properties")),
```
tobywf: I think you can probably write something to remove properties without pulling in `RegExUtils`/`commons-lang3`. I'm all for DRY, but this is a bit silly and will further bloat the classpath while the Java plugin still depends on this library.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -145,4 +148,48 @@ public static ResourceTypeSchema load(final JSONObject schemaJson) {
...
+                resourceModel));
+    }
+
+    public static void removeProperty(final JSONPointer property, final JSONObject resourceModel) {
+        String refParts = property.toString();
+        // split pointer into tokens, removing first delimiter
+        List<String> refTokens = Arrays.asList(RegExUtils.removeFirst(refParts, "/").split("/"));
+        final String key = refTokens.get(refTokens.size() - 1);
```
tobywf: think it'd be easier to make the `getRefTokens` method public?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
MattMCloudy: I can see two cases:

1. Redacting the field with a new or blank value could fail property validations.
2. Removing the filed entirely could fail validation if the writeOnly property is required.

One option could be to enforce that writeOnly properties must not also be required properties. If writeOnly properties are allowed to be required, then that means all reads of the schema are not valid representations of the resource. What do you think?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
jotompki: not used in when we strip them, both other services and packages that consume this may want to know if a model has this without wanting to strip the properties.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
jotompki: Will do good idea

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
jotompki: what are you referring to here?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
tobywf: YAGNI, we can add it later if we need it.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
tobywf: JSONPointer has a [getRefTokens](https://github.com/everit-org/json-schema/blob/2a445667317f61b28604fb7a3955fba35be49693/core/src/main/java/org/everit/json/schema/JSONPointer.java#L288-L290) method that we could use instead of the hacky approach. would have to change the visibility though, can show you how to do that.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
vladtsir: A model with redacted fields failing validation might be an overall positive side-effect - it would guarantee that the model could not be accidentally applied (for example to correct drift).

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -149,33 +148,14 @@ public static ResourceTypeSchema load(final JSONObject schemaJson) {
...
-            return property.queryFrom(resourceModel) != null;
-        } catch (JSONPointerException e) {
-            return false;
-        }
-    }
-
     public void removeWriteOnlyProperties(final JSONObject resourceModel) {
         this.getWriteOnlyPropertiesAsStrings().stream()
-            .forEach(writeOnlyProp -> removeProperty(new JSONPointer(RegExUtils.removeFirst(writeOnlyProp, "/properties")),
+            .forEach(writeOnlyProperty -> removeProperty(new PublicJSONPointer(writeOnlyProperty.replaceFirst("/properties", "")),
```
tobywf: yeah, this works great. just as a warning, `replaceFirst` takes a regex. could also anchor it or test via `startsWith`? in fact, would probably be good if `getWriteOnlyPropertiesAsStrings` just did this by default. but can iterate in future PRs

rjlohan:
My overall thinking is;

* the handler should redact its own `writeOnlyProperties` (or they are redacted by the downstream service) for `READ` 
* our contract tests should assert that `writeOnlyProperties` are _not_ returned by `READ`
* CloudFormation's "dynamic references feature for `SecretsManager` and `SSM SecureString` is a service capability that describes how CloudFormation will handle those secrets inside the engine. This capability does not overlap with the handler platform

Issue #62, opened 2019-11-27
Add support for links to resources


tobywf:
Add an e.g. `deepLink` property to the schema that allows generating links to a resource given `identifiers`. This is useful for the console.

PR #67, opened 2019-12-13
Service Link

tobywf:
*Issue #, if available:* #62 

*Description of changes:* Allow resource authors to attach template links to the schema that consumers of the schema/resources can use to generate links directly to the resource. An example might be EC2. The console link is:

```
https://eu-central-1.console.aws.amazon.com/ec2/home?region=eu-central-1#Instances:instanceId=i-xxxxxx
```

In this case, the service link might look like this:

```
"serviceLink": {
    "templateUri": "/ec2/home?region=${awsRegion}#Instances:instanceId=${instanceId}",
    "mappings": {
        "instanceId": "/InstanceId"
    }
},
```

Where `awsRegion` is a "global" variable (so far, the only one), and `instanceId` is a variable defined via a JSON pointer to a property *in a resource blob* (returned by e.g. READ) and not in the schema. The relative URI is debatable, but works well for the console currently. Other users outside of the console would have to maintain a mapping of console endpoints by region. The other benefit is that it lets us clearly figure out which links are AWS-internal.

For non-AWS services, an absolute, HTTPS URI can be used.

### DIFF
--- a/src/main/resources/examples/resource/initech.tps.report.v1.json
+++ b/src/main/resources/examples/resource/initech.tps.report.v1.json
@@ -100,7 +100,7 @@
             ]
         }
     },
-    "serviceLink": {
+    "resourceLink": {
         "$comment": "This is an example of a relative, AWS-internal service link. For external links, use an absolute URI.",
         "templateUri": "/tps/v2/home?region=${awsRegion}#Report:tpsCode=${TPSCode}",
         "mappings": {
--- a/src/main/resources/examples/resource/initech.tps.report.v1.json
+++ b/src/main/resources/examples/resource/initech.tps.report.v1.json
@@ -100,5 +100,12 @@
             ]
         }
     },
+    "serviceLink": {
+        "$comment": "This is an example of a relative, AWS-internal service link. For external links, use an absolute URI.",
+        "templateUri": "/tps/v2/home?region=${awsRegion}#Report:tpsCode=${TPSCode}",
+        "mappings": {
+            "TPSCode": "/TPSCode"
+        }
+    },
     "additionalProperties": false
 }
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -207,6 +207,33 @@
                     "additionalProperties": false
                 }
             ]
+        },
+        "serviceLink": {
+            "type": "object",
+            "properties": {
+                "$comment": {
+                    "$ref": "https://json-schema.org/draft-07/schema#/properties/$comment"
+                },
+                "templateUri": {
+                    "type": "string",
+                    "pattern": "^(/|https:)"
+                },
+                "mappings": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^[A-Za-z0-9]{1,64}$": {
+                            "type": "string",
+                            "format": "json-pointer"
+                        }
+                    },
+                    "additionalProperties": false
+                }
+            },
+            "required": [
+                "templateUri",
+                "mappings"
+            ],
+            "additionalProperties": false
         }
     },
     "type": "object",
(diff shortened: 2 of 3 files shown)


src/main/resources/examples/resource/initech.tps.report.v1.json
```diff
@@ -100,5 +100,12 @@
             ]
         }
     },
+    "serviceLink": {
```
Matt-Butler: Since a service will have several resource links, should we rename this `resourceLink`?

tobywf: think this naming came about because it's a link to a service that isn't CloudFormation. each schema is one resource type, so i'm not too worried here, besides the examples are hopefully obvious.

src/main/resources/examples/resource/initech.tps.report.v1.json
```diff
@@ -100,5 +100,12 @@
...
+        "$comment": "This is an example of a relative, AWS-internal service link. For external links, use an absolute URI.",
+        "templateUri": "/tps/v2/home?region=${awsRegion}#Report:tpsCode=${TPSCode}",
+        "mappings": {
```
rjlohan: Do we need this section? We can only pre-fill fields from the schema anyway, could the tokens in the URL just be the json-pointers directly?

tobywf: we did consider this, but this way it might be possible to use existing templating engines, or simple regex replace. might be good for the console team to create a PoC to see how difficult/easy both would be.

rjlohan: Fair enough.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -207,6 +207,33 @@
...
+                            "type": "string",
+                            "format": "json-pointer"
+                        }
+                    },
+                    "additionalProperties": false
+                }
+            },
+            "required": [
+                "templateUri",
+                "mappings"
```
PatMyron: `mappings` could be optional, right?

src/main/resources/examples/resource/initech.tps.report.v1.json
PatMyron: [`AWS::DAX::ParameterGroup` ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dax-parametergroup.html) is a good candidate for a proof-of-concept:

`/dynamodb/home#cache-parameter-group:selected=${lowercasedResourceId}`

benkehoe:
Does this support other partitions, like China? Additionally, what if there is more than one kind of link that a provider wants to define?

src/main/resources/schema/provider.definition.schema.v1.json
tobywf: i don't see how that could produce sensible links to the resource, since you always have to map some kind of information into to the link. and to keep the templating operation simple, you have to map information to variables (you can't put a JSON pointer in the template URI). i guess you could define an empty `mappings`, but this way there's one less check required for consuming code to see if `mappings` is even present.

tobywf:
Yes, partitions are not an issue for AWS links as they have to be relative. For external links, they can go off the region. If there's really a need for the partition, we can add that as a global variable later.

1:1 seems to make sense to keep things simple. I don't think we're keen on displaying multiple links in the stack resources tab either, could lead to a confusing UX. I'm also not aware of any use-cases for multiple links. And while that's a somewhat weak argument as people do build cool stuff with extensible systems, predicting use-cases is hard and I'm leaning towards simplicity.

src/main/resources/examples/resource/initech.tps.report.v1.json
tobywf: changed my mind on this, think you're right and resource link is more natural. fixed!

tobywf:
Happy with the state of this and think it's ready for inclusion!

PatMyron:
@Matt-Butler and fallback on the old `deepLinkResourceSpecification.json` file from https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/139#issuecomment-536230778 so we don't lose support for resource types we already support?

PatMyron:
only concern is the discoverability of things like `${awsRegion}` if more get supported, but we can improve that while adding more

PR #68, opened 2019-12-14
Adjust auto-formatter settings to not join lines and re-flow comments

vladtsir:
*Issue #, if available:*

*Description of changes:*
I propose that automatic formatter should not  reflows comments and join lines together.

If I write something like this:
```
Schema = Schema.builder()
        .withDraftV7()
        .withMetaSchema(schema1)
        .withMetaSchema(schema2)
        .withSomeOtherOption(true)
        .build();
```
I do not want it to be changed to 
```
Schema = Schema.builder().withDraftV7().withMetaSchema(schema1).withMetaSchema(schema2).withSomeOtherOption(true).build();
```

The same goes for comments. I usually put some thought into where the line breaks should go in the comments; formatter usually makes it worse and never makes it better

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -42,103 +42,134 @@ public class ResourceTypeSchema extends ObjectSchema {
     private final List<JSONPointer> createOnlyProperties = new ArrayList<>();
     private final List<JSONPointer> deprecatedProperties = new ArrayList<>();
     private final List<JSONPointer> primaryIdentifier = new ArrayList<>();
-    private final List<List<JSONPointer>> additionalIdentifiers = new ArrayList<>();
+    private final List<
+        List<JSONPointer>> additionalIdentifiers = new ArrayList<>();
     private final List<JSONPointer> readOnlyProperties = new ArrayList<>();
     private final List<JSONPointer> writeOnlyProperties = new ArrayList<>();
 
     public ResourceTypeSchema(final ObjectSchema.Builder builder) {
         super(builder);
 
-        super.getUnprocessedProperties().forEach(this.unprocessedProperties::put);
+        super.getUnprocessedProperties()
+            .forEach(this.unprocessedProperties::put);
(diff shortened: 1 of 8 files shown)


rjlohan:
Yeah I think leave the line length at 80, the rest is OK.

vladtsir:
> Yeah I think leave the line length at 80, the rest is OK.

this was for the width of comments. max line width was not changed in this CR - had been at 130 all along - ln 311:
```
<setting id="org.eclipse.jdt.core.formatter.lineSplit" value="130"/>
```
If I set both to 80 then 7 classes will be re-formatted:
```
src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
src/main/java/software/amazon/cloudformation/resource/SchemaValidator.java
src/main/java/software/amazon/cloudformation/resource/Validator.java
src/main/java/software/amazon/cloudformation/resource/exceptions/ValidationException.java
src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
src/test/java/software/amazon/cloudformation/resource/exceptions/ValidationExceptionTest.java
```
should I do it?

rjlohan:
Post the diff as an extra commit and we'll see what it looks like?

tobywf:
sorry, didn't realize it was only for comments - can set that to 130, too then. either way is fine.

vladtsir:
I think 80 is a bit narrow. What if I want to have more than 15 nested loops?

rjlohan:
> I think 80 is a bit narrow. What if I want to have more than 15 nested loops?

You have more problems than line width at that point... but yeah 80 chars does create some crappy line breaks in that diff. @tobywf what do you think?

tobywf:
happy to revert the last two and leave as is for now. don't want to get bogged down in this, and it's already at 130. thanks to the auto-formatter, can change it whenever we want in future anyway.

it's a balance. 80 is defo short, and can feel pointless on a huge screen. at the same time, you and me do a lot of reviews in meetings where we don't have external monitors. but it seems to have been working fine so far.

Issue #23, opened 2019-07-09
Validate $ref correctly


tobywf:
# Issue

See also https://github.com/aws-cloudformation/aws-cloudformation-rpdk/issues/66

TL;DR: If we validate the schema as data, and our meta schema allows `$ref`, it doesn't work, as the "data" i.e. the schema is not further processed:

```java
@Test
public void improperValidation() {
    final JSONObject definition = new JSONObject()
            .put(TYPE_NAME_KEY, EXAMPLE_TYPE_NAME)
            .put(DESCRIPTION_KEY, EXAMPLE_DESCRIPTION)
            .put(PROPERTIES_KEY, new JSONObject()
                .put("propertyA", new JSONObject()
                    .put("$ref", "#/garbage")))
            .put(PRIMARY_IDENTIFIER_KEY, Arrays.asList(EXAMPLE_PRIMARY_IDENTIFIER));

    validator.validateResourceDefinition(definition);
}
```

this passed, but the `$ref` is obviously garbage. The reason it "passes" is because we allow `$ref`s:

https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/6a1ef45d612d7c90da1d77e71a9427a64cdd5237/src/main/resources/schema/provider.definition.schema.v1.json#L85-L87

however, all JSON schema mandates is:

```json
"$ref": {
    "type": "string",
    "format": "uri-reference"
},
```

# Solutions?

The obvious solution is to resolve all `$ref`s before validation, and then it'd work. In Python, I'd use [jsonref](https://pypi.org/project/jsonref/), or disallow circular references and simply inline all references, and disallow `$ref`.

It may be possible to use [org.everit.json.schema.loader.SchemaLoader](https://github.com/everit-org/json-schema/blob/0d797d03bafccb7d6f050ba4bb20758f65d1eafb/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java) in some capacity. Instead of validating against draft-7 (the usual meta-schema), it may be possible to use our meta-schema instead. Since [`SchemaLoader` handles `$ref`s semantically](https://github.com/everit-org/json-schema/blob/0d797d03bafccb7d6f050ba4bb20758f65d1eafb/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java#L401), that ought to work. (I had planned to do something like this in the RPDK, where the library/dynamic nature makes it fairly straightforward to do.)

[...]

PR #69, opened 2019-12-15
Validate ref

vladtsir:
*Issue #, if available:*
#23  
*Description of changes:*

Everit validates $ref values, when a Schema is created using SchemaLoader$load(). 

This change adds loading resource definition as a schema with ResourceDefinitionSchema and JSON schema registered as meta-schemas. Unit tests show that invalid references cause SchemaLoader to throw a ValidationException.


tobywf:
this shares history with #68 - need to get that merged first. i've already closed #72 , because three dependent PRs is definitely too much. if you want to leave this one open for review, you'll have to get #68 merged first, and then merge `master` into this PR. do not rebase if people have already reviewed, it breaks links between comments and code because history is rewritten on rebase.

src/main/resources/schema/schema
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft-07/schema#",
-  "$id": "https://json-schema.org/draft-07/schema#",
+  "$id": "https://json-schema.org/draft-07/schema",
```
vladtsir: `Validator` uses `$id` value as the key when caching meta-schemas (see ln159). The trailing # throws off the resolver.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://json-schema.org/draft-07/schema#",
-    "$id": "provider.definition.schema.v1.json",
+    "$id": "https://schema.cloudformation.us-east-1.amazonaws.com/provider.definition.schema.v1.json",
```
vladtsir: exact value of the $id doesn't matter here, assuming we do not plan to deploy different provider definitions in different regions.

pom.xml
```diff
@@ -56,11 +56,32 @@
...
+            <version>5.1.1</version>
+            <scope>test</scope>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter-engine -->
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter</artifactId>
-            <version>5.5.1</version>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>5.1.1</version>
```
vladtsir: Downgraded jupiter to 5.1.1 because 5.5.1 is not compatible with mockito 2.22.0

rjlohan: Yeah can we upgrade the mockito dependency rather than downgrade something else?

src/main/java/software/amazon/cloudformation/resource/Validator.java
```diff
@@ -20,43 +20,64 @@
...
 import software.amazon.cloudformation.resource.exceptions.ValidationException;
 
 public class Validator implements SchemaValidator {
 
-    private static final String JSON_SCHEMA_ID = "https://json-schema.org/draft-07/schema";
+    private static final String ID_KEY = "$id";
     private static final String JSON_SCHEMA_PATH = "/schema/schema";
-    private static final String METASCHEMA_PATH = "/schema/provider.definition.schema.v1.json";
+    private static final String RESOURCE_DEFINITION_SCHEMA_PATH = "/schema/provider.definition.schema.v1.json";
+    /**
```
rjlohan: nit: can we line-break before new comment blocks?

src/main/resources/schema/provider.definition.schema.v1.json
rjlohan: It's possible to have regional discrepancies in a type. e.g; rolling out a new feature would mean that there's a period of time where the type is different in different regions. Also, there are occasionally longer lags on feature support into new regions that are expressed as properties on a resource. In that case, we'd see different definitions for a type in each region.

src/test/java/software/amazon/cloudformation/resource/ValidatorRefResolutionTests.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.verification.VerificationMode;
+
+import software.amazon.cloudformation.resource.exceptions.ValidationException;
+
+/**
+ *
```
rjlohan: Don't need empty comment block.

pom.xml
tobywf: i think we can leave it for now. it's only a test dependency

src/main/resources/schema/provider.definition.schema.v1.json
vladtsir: rjlohan@ -  I think you are asking about type definitions. If so, the answer is yes, I do expect that there will be regional discrepancies - some of the types will not be available in some regions. For type definitions we will inject different $id values in each region; this way types using relative references will be validated correctly in each region. We do not have to do this for the resource definition schema because:
- resource providers do not have to reference resource definition schema at all - all our tools will validate against this schema regardless and will ignore any $schema values supplied in the resource definition
- if we release a new version of the schema it will be hosted at a different URI i.e. `provider.definition.schema.v2.json`. I (and all our tools) assume that we will not make the new definition schema public until its fully supported in all regions - exactly because all resource definitions are validated against the same schema.

Having said that, I did not have to change the $id of provider.definition.schema.v1.json at all. I think I did that because I was cross-testing schema changes using `ajv` which needs an explicit `$schema` key to function. I will probably revert this in the next commit

rjlohan: Ah right, misunderstood this change.

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -418,4 +412,42 @@ public void validateExample_exampleResource_shouldBeValid() throws IOException {
...
+     */
+    @ParameterizedTest
+    @ValueSource(strings = { ":invalid/uri", "" })
+    public void registerMetaSchema_invalidRelativeRef_shouldThrow(String uri) {
+
+        JSONObject badSchema = loadJSON(RESOURCE_DEFINITION_SCHEMA_PATH);
+        badSchema.put("$id", uri);
+        assertThatExceptionOfType(ValidationException.class).isThrownBy(() -> {
+            validator.registerMetaSchema(SchemaLoader.builder(), badSchema);
+        });
```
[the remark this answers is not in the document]
vladtsir: as luck would have it this test was removed in the next commit - I am no longer extracting using the schema $id as URI, so no need to test cases when $id is missing.

src/test/java/software/amazon/cloudformation/resource/ValidatorRefResolutionTests.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    private JSONObject getSampleTemplate() {
+        final JSONObject template = new JSONObject();
+        template.put("Time", "2019-12-12T10:10:22.212Z");
+        return template;
+    }
+
+    private static VerificationMode twice() {
+        return Mockito.times(2);
+    }
```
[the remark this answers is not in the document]
vladtsir: I think `verify(downloader, twice()).get(expectedRefUrl);` reads better, esp in intelliJ. Many of us have "show parameter name hints" enabled, in which case intelliJ shows something like

`verify(downloader, times( wantedNumberOfInvocations: 2)).get(expectedRefUrl);`

It's configurable though (IntelliJ). Will remove the method

src/main/java/software/amazon/cloudformation/resource/Validator.java
```diff
@@ -68,7 +95,72 @@ public void validateObject(final JSONObject modelObject, final JSONObject defini
...
+        // inject/replace $schema URI to ensure that provider definition schema is used
+        definition.put("$schema", RESOURCE_DEFINITION_SCHEMA_URI.toString());
         validateObject(definition, definitionSchemaJsonObject);
+        // validateObject cannot validate schema-specific attributes. For example if definition
+        // contains "propertyA": { "$ref":"./some-non-existent-location.json#definitions/PropertyX"}
+        // validateObject will succeed, because all it cares about is that "$ref" is a URI
+        // In order to validate that $ref points at an existing location in an existing document
+        // we have to "load" the schema
+        loadResourceSchema(definition);
     }
```
jotompki: is there benefit in waiting to validate the object until after the references are loaded, say the customer references something for an object that requires one property, but the referenced object doesnt have it?

tobywf: when i was playing around with this, some constraints in the meta schema weren't being validated when the schema was loaded. conversely, because of the `$ref` is type string issue, validating those wasn't happening in the first step. put together, hopefully this gets both. it's of course possible to ref something that isn't valid, especially since right now, refs aren't restricted. in future, they should be restricted to same-document or relative refs, so then we always know they're referring to another hosted schema (and regionalisation works when we inject `$id`). based on that constraint, it's safe to assume the other schemas are valid, so hopefully there's no issue.

of course, deeply validating the schema would be ideal. but this is already an improvement.

src/main/java/software/amazon/cloudformation/resource/Validator.java
```diff
@@ -68,7 +95,72 @@ public void validateObject(final JSONObject modelObject, final JSONObject defini
...
+     */
+    public Schema.Builder<?> getResourceSchemaBuilder(final JSONObject resourceDefinition) {
+        final SchemaLoaderBuilder loaderBuilder = getSchemaLoader(resourceDefinition);
+        loaderBuilder.registerSchemaByURI(RESOURCE_DEFINITION_SCHEMA_URI, definitionSchemaJsonObject);
+
+        final SchemaLoader loader = loaderBuilder.build();
+        try {
+            return loader.load();
+        } catch (org.everit.json.schema.SchemaException e) {
+            throw new ValidationException(e.getMessage(), e.getSchemaLocation(), e);
```
tobywf: could we convert the `SchemaException` to a `org.everit.json.schema.ValidationException` and keep using `causingExceptions` instead?

src/main/java/software/amazon/cloudformation/resource/Validator.java
jotompki: makes sense definitely something we can add down the line

Issue #70, opened 2019-12-15
ResourceTypeSchema fails when loading resource definitions using 'oneOf', 'if' keywords


vladtsir:
Error:
```
java.lang.ClassCastException: org.everit.json.schema.CombinedSchema$Builder cannot be cast to org.everit.json.schema.ObjectSchema$Builder
	at software.amazon.cloudformation.resource.ResourceTypeSchema.load(ResourceTypeSchema.java:110)
```

To reproduce create a schema file as follows:

*valid-with-oneof.json*:

```
{
  "typeName": "AWS::Test::TestModel",
  "description": "A test schema for unit tests.",
  "sourceUrl": "https://mycorp.com/my-repo.git",
  "properties": {
    "propertyA": {
      "type": "string"
    },
    "propertyB": {
      "type": "string"
    }
  },
  "oneOf": [
    { "required":  ["propertyA"]},
    { "required" : ["propertyB"]}
  ],
  "primaryIdentifier": [
    "/properties/propertyA"
  ],
  "additionalProperties": false
}
```

```
public void validSchema_withOneOf_shouldSucceed() {
        JSONObject resource = loadJSON("/valid-with-oneof.json");
        final ResourceTypeSchema schema = ResourceTypeSchema.load(resource);
}
```

PatMyron:
@vladtsir does https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/81 fix this?

PR #71, opened 2019-12-16
Fix oneof schema

vladtsir:
*Issue #, if available:*
#70 

*Description of changes:*

Looking at how `ResourceTypeSchema` is used, I'm rather convinced that it should not subclass from `ObjectSchema` (or `Schema`) at all. Its primary use is to provide strongly-typed access to schema elements; this can be done if the schema is contained within. Currently nothing uses ResourceTypeSchema to validate actual CFN templates, but if necessary, this can be easily done by forwarding "validate" call to the `schema` member


vladtsir:
It appears that I inadvertently include two of my other pull requests in this one...

PR #86, opened 2020-03-13
Allow replacementStrategy in the schema

wenyhu:
*Description of changes:* Allow replacement Strategy in the schema. This allows immutable singleton resources to update in the order of [delete, create]. 

Valid strategies are 
**- [create, delete]**
**- [delete, create]**

Right now we always try to create then delete for immutable updates, allowing this replacementStrategy in the schema would allow update behavior be decided by the customer. Most useful singleton update that can have only one instance. 

There should be be NO change in behavior for current customers as not specifying a strategy would default to [create, delete], which is how it behaves now.


ikben:
Out of curiosity.

Since this is a setting that's supplied by the provider, are there any plans to allow a consumer to restrict the usage of this strategy? Currently CloudFormation consumers assume that every resource will only be deleted if all creates succeed.

Eg. by having a capability CAPABILITY_DELETE_BEFORE_CREATE.

On the CloudFormation side (independent of the schema), it might make sense to do these resources as late as possible so that if another resource fails the delete hasn't happend yet

wenyhu:
> Out of curiosity.
> 
> Since this is a setting that's supplied by the provider, are there any plans to allow a consumer to restrict the usage of this strategy? Currently CloudFormation consumers assume that every resource will only be deleted if all creates succeed.
> 
> Eg. by having a capability CAPABILITY_DELETE_BEFORE_CREATE.
> 
> On the CloudFormation Engine side (independent of the schema), it might make sense to do these resources as late as possible so that if another resource fails the delete hasn't happened yet

Makes sense. So if a customer do not specifically opt in for delete first by adding a replacementStrategy, we will make the [create,delete] default and not change their current behavior (create then delete). Let me discuss further with the team on restrictions on this and loop back. Thanks for the helpful suggestion!

ikben:
> So if a customer do not specifically opt in for delete first by adding a replacementStrategy, we will make the [create,delete] default and not change their current behavior (create then delete)

In this case you mean the customer when he's writing the resource, right?

It might makes sense to differentiate between the customer writing the resource (who explicitly changes the spec to override the default) and the customer using the resource (who can add a capability to a action). In a lot of cases this will be different people.

aygold92:
here we are talking about the behavior of the resource itself.  For singleton resources (resources where only 1 of them can exist), we obviously cannot take the normal strategy of create first, then delete, since we cannot create a second resource.  

There are some resources that CFN currently supports that behave this way, such as `AWS::OpsWorks::ElasticLoadBalancerAttachment`.  During an update, CFN will delete and then create.

In order to have consistent behavior, we want to make the contract more explicit for Uluru, so we do not allow Update handlers to create a new physical resource.  Introducing a value into the schema would allow us to more predictably and consistently support this use case.

ikben:
> There are some resources that CFN currently supports that behave this way, such as AWS::OpsWorks::ElasticLoadBalancerAttachment. During an update, CFN will delete and then create
Interesting! I will need to start adding more footnotes to my "this is how CFN works" explanations.

> In order to have consistent behavior, we want to make the contract more explicit for Uluru, so we do not allow Update handlers to create a new physical resource. Introducing a value into the schema would allow us to more predictably and consistently support this use case

To be clear, I think that's a great idea. I asked mostly out of interests of how this would manifest to the end (template) user, who might - like me, wrongly - assume that a delete never happens in the update in progress phase.

> The stack would then start rolling back and would eventually create the old resource again, but it might then be while before it gets to that point

Yes, and since third party providers use the same schema you have no guarantee that the old resource is recreatable (it should be, but people have a tendency to aim at their own feet sometimes).

> That being said, I don't see any way around this risk. How else would they do it? What would be the behavior of the resource if they didn't opt in?
If it's a real singleton, then there is no way around this. Besides not updating / replacing it.
If it's not a real singleton they can open a bug with the resource provider (or do the two steps themselves in two updates).

However you can still reduce the risk, eg by doing the update of the singleton (if possible) in its own operation, and everything else in another.T

his is similar to how other capabilities work. I either accept the risk that the template developer can create IAM resources or I don't deploy the resources. And I can similarly read the policies in the template to reduce the risk of there being something dangerous in there.

That being said, I'm not certain that an extra capability is the best solution here (especially if there are already resources with this behaviour), but i do think consumer might want to know when a provider behaves unintuitive.

I was also thinking about what happens if the provider throws an NotUpdatable exception. But since I haven't tested the current behaviour of that, I'm assuming that it will do the right thing in both cases.


Thanks for the thoughtful responses!

rjlohan:
I like the idea that we add a way for customers to block stack changes that do these non-standard replacement changes. That wasn't originally in scope, but it's a good idea. I think we can move ahead with the declarative changes as a start and then look at how to enable the protection idea.

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
```diff
@@ -101,6 +101,16 @@ public void getWriteOnlyProperties() {
         assertThat(result).containsExactly("/properties/propertyC", "/properties/propertyE/nestedProperty");
     }
 
+    @Test
+    public void getReplacementStrategy() {
```
jotompki: can we have a test that tests the opposite [delete, create] as well? just would be good for sanity to know that this deserializes correctly in both cases.

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
```diff
@@ -101,6 +101,16 @@ public void getWriteOnlyProperties() {
...
+        JSONObject o = loadJSON(TEST_SCHEMA_PATH);
+        final ResourceTypeSchema schema = ResourceTypeSchema.load(o);
+
+        List<String> result = schema.getReplacementStrategy();
+        assertThat(result).containsExactly("create", "delete");
```
jotompki: can we have a test for the reverse as well? might be good for sanity to know both cases serialize correctly

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -75,6 +76,11 @@ public ResourceTypeSchema(Schema schema) {
             : null;
         this.unprocessedProperties.remove("$schema");
 
+        this.unprocessedProperties.computeIfPresent("replacementStrategy", (k, v) -> {
+            ((ArrayList<?>) v).forEach(p -> this.replacementStrategy.add(p.toString()));
+            return null;
```
jotompki: are we not going to provide the default value here if they aren't present? ["create", "delete"]

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
wenyhu: 👌will add

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
wenyhu: Ah I was thinking about handling in the workflows where if a schema does not have replacementStrategy info will set it to [create, delete]. But doing it here seems neater to me. 

Is there like a convention on how not present variables are handled?

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
wenyhu: > can we have a test that tests the opposite [delete, create] as well? just would be good for sanity to know that this deserializes correctly in both cases.
👌will add

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -322,6 +322,45 @@ public void validateDefinition_invalidHandlerSection_shouldThrow() {
...
+            .withMessageContaining("#/replacementStrategy");
+    }
+
+    @Test
+    public void validateDefinition_InvalidReplacementStrategy_shouldThrow() {
+        List<String> invalidReplacementStrategy = Arrays.asList("create", "update");
+        final JSONObject definition = baseSchema().put("replacementStrategy", invalidReplacementStrategy);
+        assertThatExceptionOfType(ValidationException.class).isThrownBy(() -> validator.validateResourceDefinition(definition))
+            .withMessageContaining("#/replacementStrategy");
+    }
```
aygold92: can we parameterize this to make it 1 test? (see `@ParameterizedTest` elsewhere in the class)

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -322,6 +322,45 @@ public void validateDefinition_invalidHandlerSection_shouldThrow() {
             .withNoCause().withMessage("#/handlers/read: required key [permissions] not found");
     }
 
+    @Test
+    public void validateDefinition_validReplacementStrategy_shouldNotThrow() {
```
aygold92: parameterize this one too to include both create/delete and delete/create

wenyhu: Good point. Parameterized all the tests

wenyhu:
> can you add a section to the README that explains this property?

Will add.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -292,6 +305,10 @@
             ],
             "$ref": "#/definitions/httpsUrl"
         },
+        "replacementStrategy": {
+            "$comment": "",
```
jotompki: Should add comment here

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -71,6 +71,19 @@
                 }
             }
         },
+        "replacementStrategy": {
+            "type": "array",
```
jotompki: description might be good to add as well as [default value](https://json-schema.org/understanding-json-schema/reference/generic.html) since we are defaulting to something.

wenyhu: Good idea. Will add description and default value

src/main/resources/schema/provider.definition.schema.v1.json
wenyhu: Oops forgot about that. Will add.

README.md
```diff
@@ -35,6 +35,7 @@ Certain properties of a resource are _semantic_ and have special meaning when us
...
 * **`deprecatedProperties`**: A property in the `deprecatedProperties` is not guaranteed to be present in the response from a **READ** request. These fields will still be accepted as input to **CREATE** and **UPDATE** requests however they may be ignored, or converted to new API forms when outbound service calls are made.
+* **`replacementStrategy`**: Used for resource that needs replacement upon an update, where CloudFormation recreates the resource and generates a new physical ID. Must be either `[create, delete]` or `[delete, create]`. Default value is `[create, delete]`
```
aygold92: I think the description here is a bit confusing, and we want to give a bit more context.  Let me give it a try...

"""
As mentioned above, changing a `createOnlyProperty` requires replacement of the resource by creating a new one and deleting the old one.  The default CloudFormation replacement behavior is to create a new resource first, then delete the old resource, so as to avoid any downtime. However, some resources are singleton resources, meaning that only one can exist at a time. In this case, it is not possible to create a second resource first, so CloudFormation must Delete first and then Create.  Specify either `[create, delete]` or `[delete, create]`. Default value is `[create, delete]`
"""

Now that I'm writing that, this does seem to be bleeding into CloudFormation specific behavior.  Are we okay with that @rjlohan?  

I think the bigger concern here is that without the context of CloudFormation, this doesn't really mean anything.  Someone might look at their resource and say "I dunno, it doesn't really matter whether you create then delete or delete then create. You could do it either way".  This lends me to think maybe this property should take some other form.

That's why I like the idea of `isSingletonResource` or some other boolean value that is not open to interpretation.

rjlohan: I think this is OK, and it's not really a CloudFormation behaviour, it's a function of how a resource can be safely replaced which I think is more general. I also prefer that we use some explicit naming rather than a flag and name that has to be interpreted. `isSingletonResource` doesn't tell me anything about what this field does at runtime. I think the `replacementStrategy` naming is clearer here, and the values are more expicit.

rjlohan: I'm OK with the text update though. 👍

rjlohan:
@vladtsir has a good suggestion. As I commented earlier, I like the naming of this property as it expresses the impact of the setting. However it doesn't need to be a list. A string (scalar) with enum does the same job with less hassle.


```
"replacementStrategy": {
            "type": "string",
            "description": "The valid replacement strategies are [create_then_delete] and [delete_then_create]. All other inputs are invalid.",
            "default": "create_then_delete",
            "enum": [
                "create_then_delete",
                "delete_then_create"
            ]
        },
```

There's only one other value I think might come up eventually which would be something like 'in place' - where an Update action triggers the replacement. An API that offers only a HTTP PUT on a List _might_ be such a thing.

rjlohan (approved):

rjlohan (approved):

PR #103, opened 2020-10-15
Add metadata flags for drift

anshikg:
*Issue #, if available:*

*Description of changes:* Add new metadata flags for Uluru schemas to make the schema more verbose for drift.

*Test:*
* build the package
* added unit test for new metadata flags 

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 #### optionalPrefix/optionalSuffix
 
 Some properties can be updated by the downstream service teams and these properties could have a prefix/suffix or both once returned from the Read Handler. These metadata properties are strings
-which can be provided for a property and checked on drift to ensure the property is not drifted.
+which can be provided for a property and checked on drift to ensure the property has not drifted.
 
 `optionalPrefix: "anyString"`
 `optionalSuffix: '"anyString"`
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ which can be provided for a property and checked on drift to ensure the property
 
 #### nullEqualsEmpty
 
-Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a boolean true|false.
 By default drift considers null and empty list to be different.
 `nullEqualsEmpty: true`
 
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 
 #### optionalPrefix/optionalSuffix
 
-Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler. These metadata properties are strings
+Some properties can be updated by the downstream service teams and these properties could have a prefix/suffix or both once returned from the Read Handler. These metadata properties are strings
 which can be provided for a property and checked on drift to ensure the property is not drifted.
 
 `optionalPrefix: "anyString"`
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ By default drift considers null and empty list to be different.
 
 #### caseSensitive
 
-Some service teams support properties which are case insensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
+Some service teams support properties that are case insensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a boolean true|false.
 By default drift is case sensitive.
 `caseSensitive: true`
 
--- a/README.md
+++ b/README.md
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 | true           | true           | ordered set    |
 | false          | true           | set      |
 
+#### optionPrefix/Suffix
+
+Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler. These metadata properties are strings
+which can be provided for a property and checked on drift to ensure the property is not drifted.
+
+`optionalPrefix: "anyString"`
+`optionalSuffix: '"anyString"`
+
+#### nullEqualsEmpty
+
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
+By default drift considers null and empty list to be different.
+`nullEqualsEmpty: true`
+
+#### caseInsesitive
+
+Some service teams support properties which are caseInsensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
+By default drift is case sensitive.
+`caseInsesitive: true`
 
 ### Constraints
 
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 | true           | true           | ordered set    |
 | false          | true           | set      |
 
-#### optionPrefix/Suffix
+#### optionalPrefix/optionalSuffix
 
 Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler. These metadata properties are strings
 which can be provided for a property and checked on drift to ensure the property is not drifted.
@@ -153,11 +153,11 @@ Some service teams consider null and empty list to be the same. This metadata pr
 By default drift considers null and empty list to be different.
 `nullEqualsEmpty: true`
 
-#### caseInsesitive
+#### caseSensitive
 
-Some service teams support properties which are caseInsensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
+Some service teams support properties which are case insensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
 By default drift is case sensitive.
-`caseInsesitive: true`
+`caseSensitive: true`
 
 ### Constraints
 
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -85,6 +85,26 @@
                             "type": "boolean",
                             "default": true
                         },
+                        "nullEqualsEmpty": {
+                            "description": "When set to false, this flag indicates that null and empty list are not equal, and changing the values from null to empty list would indicate diff",
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "caseInsensitive": {
+                            "description": "When set to false, this flag indicates that case of the string will be honored, and changing the case of the string would indicate diff",
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "optionalSuffix": {
+                            "description": "When set, this flag indicates that the read handler will return this property with suffixed string",
+                            "type": "string",
+                            "pattern": "^[a-zA-Z0-9$&+,:?@#|<>.^*()%!-]{0,}"
+                        },
+                        "optionalPrefix": {
+                            "description": "When set, this flag indicates that the read handler will return this property with prefixed string",
+                            "type": "string",
+                            "pattern": "^[a-zA-Z0-9$&+,:?@#|<>.^*()%!-]{0,}"
+                        },
                         "$ref": {
                             "$ref": "http://json-schema.org/draft-07/schema#/properties/$ref"
                         },
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -90,10 +90,10 @@
                             "type": "boolean",
                             "default": false
                         },
-                        "caseInsensitive": {
-                            "description": "When set to false, this flag indicates that case of the string will be honored, and changing the case of the string would indicate diff",
+                        "caseSensitive": {
+                            "description": "When set to true, this flag indicates that case of the string will be honored, and changing the case of the string would indicate diff",
                             "type": "boolean",
-                            "default": false
+                            "default": true
                         },
                         "optionalSuffix": {
                             "description": "When set, this flag indicates that the read handler will return this property with suffixed string",
--- a/src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
+++ b/src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
@@ -39,6 +39,7 @@ import software.amazon.cloudformation.resource.exceptions.ValidationException;
 public class ValidatorTest {
     private static final String RESOURCE_DEFINITION_SCHEMA_PATH = "/schema/provider.definition.schema.v1.json";
     private static final String TEST_SCHEMA_PATH = "/test-schema.json";
+    private static final String SPECIAL_SCHEMA_PATH = "/special-schema.json";
     private static final String TEST_VALUE_SCHEMA_PATH = "/scrubbed-values-schema.json";
     private static final String SCHEMA_WITH_HANDLERS_PATH = "/valid-with-handlers-schema.json";
     private static final String TYPE_NAME_KEY = "typeName";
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+`optionalPrefix: "anyString"`
+`optionalSuffix: '"anyString"`
+
+#### nullEqualsEmpty
+
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
+By default drift considers null and empty list to be different.
+`nullEqualsEmpty: true`
+
+#### caseInsesitive
```
PatMyron, line 156: typo
```suggestion
#### caseInsensitive
```

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -85,6 +85,26 @@
                             "type": "boolean",
                             "default": true
                         },
+                        "nullEqualsEmpty": {
+                            "description": "When set to false, this flag indicates that null and empty list are not equal, and changing the values from null to empty list would indicate diff",
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "caseInsensitive": {
```
PatMyron, line 93: thoughts on reversing this to `caseSensitive` instead? Might be more straightforward than specifying something as not `caseInsensitive`
```suggestion
                        "caseSensitive": {
```

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 | true           | true           | ordered set    |
 | false          | true           | set      |
 
+#### optionPrefix/Suffix
```
PatMyron, line 142: just in case anyone is searching for documentation
```suggestion
#### optionalPrefix/optionalSuffix
```

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
+By default drift considers null and empty list to be different.
+`nullEqualsEmpty: true`
+
+#### caseInsesitive
+
+Some service teams support properties which are caseInsensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
+By default drift is case sensitive.
+`caseInsesitive: true`
```
PatMyron, line 160: typo
```suggestion
`caseInsensitive: true`
```

PatMyron:
Something unclear about these: are these specifically about `ReadHandlers`? For example, [`AWS::ImageBuilder::Component.Platform`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-imagebuilder-component.html#cfn-imagebuilder-component-platform) must be input with the casing `Linux`, but the `ReadHandler` returns `LINUX` for that value currently. Is this property case sensitive or case insensitive?

src/main/resources/schema/provider.definition.schema.v1.json
anshikg: valid point

README.md
anshikg: sure

anshikg:
> Something unclear about these: are these specifically about `ReadHandlers`? For example, currently [`AWS::ImageBuilder::Component.Platform`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-imagebuilder-component.html#cfn-imagebuilder-component-platform) must be input with the casing `Linux`, but the `ReadHandler` returns `LINUX` for that value. Is that property case sensitive or case insensitive? Should we consider an `enum` instead of a `boolean`?

According to the above implementation if the team does not care wether the value is Linux/LINUX they case set this to the caseInsensitive flag to true. Enum is probably a good route if the property have finite set of values. In case, the property is accepting any values use the case sensitive flag is good route. 
Another thing, in the above case we just need to be careful we don't break the existing stacks. If the customer was providing Linux/LINUX as input going forward both should be acceptable

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+
+#### nullEqualsEmpty
+
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
+By default drift considers null and empty list to be different.
+`nullEqualsEmpty: true`
+
+#### caseSensitive
+
+Some service teams support properties which are case insensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a bollean true|false.
```
ikben, line 158:
```suggestion
Some service teams support properties that are case insensitive. This metadata property allows such properties to avoid drift when ignoring case. This metadata takes in a boolean true|false.
```

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+
+Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler. These metadata properties are strings
+which can be provided for a property and checked on drift to ensure the property is not drifted.
+
+`optionalPrefix: "anyString"`
+`optionalSuffix: '"anyString"`
+
+#### nullEqualsEmpty
+
+Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a bollean true|false.
```
ikben, line 152:
```suggestion
Some service teams consider null and empty list to be the same. This metadata property allows such properties to avoid drift when they equating null = []. This metadata takes in a boolean true|false.
```

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -85,6 +85,26 @@
...
+                        "nullEqualsEmpty": {
+                            "description": "When set to false, this flag indicates that null and empty list are not equal, and changing the values from null to empty list would indicate diff",
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "caseSensitive": {
+                            "description": "When set to true, this flag indicates that case of the string will be honored, and changing the case of the string would indicate diff",
+                            "type": "boolean",
+                            "default": true
+                        },
```
PatMyron, line 97: Don't think we have enough data to make an informed decision to choose these default values yet. Can we postpone specifying the default value to encourage explicit schematization for now?

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -85,6 +85,26 @@
...
+                        "optionalSuffix": {
+                            "description": "When set, this flag indicates that the read handler will return this property with suffixed string",
```
jotompki, line 99: flag isn't the correct terminology here imo. Flag usually indicates a boolean value whereas this is just a string

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -85,6 +85,26 @@
...
+                            "type": "string",
+                            "pattern": "^[a-zA-Z0-9$&+,:?@#|<>.^*()%!-]{0,}"
+                        },
+                        "optionalPrefix": {
+                            "description": "When set, this flag indicates that the read handler will return this property with prefixed string",
```
jotompki, line 104: same here. are we planning on handling this sort of prefix additions handler side? if so would have to do for every language plugin

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
```diff
@@ -39,6 +39,7 @@
 public class ValidatorTest {
     private static final String RESOURCE_DEFINITION_SCHEMA_PATH = "/schema/provider.definition.schema.v1.json";
     private static final String TEST_SCHEMA_PATH = "/test-schema.json";
+    private static final String SPECIAL_SCHEMA_PATH = "/special-schema.json";
```
jotompki, line 42: can we use a different name? special isn't super descriptive

jotompki (changes requested):

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 | true           | true           | ordered set    |
 | false          | true           | set      |
 
+#### optionalPrefix/optionalSuffix
+
+Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler. These metadata properties are strings
```
jotompki, line 144:
```suggestion
Some properties can be updated by the downstream service teams and these properties could have a prefix/suffix or both once returned from the Read Handler. These metadata properties are strings
```

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
...
+which can be provided for a property and checked on drift to ensure the property is not drifted.
```
jotompki, line 145:
```suggestion
which can be provided for a property and checked on drift to ensure the property has not drifted.
```

ammokhov (approved):

src/main/resources/schema/provider.definition.schema.v1.json
anshikg: would metadata work better than flag?

src/main/resources/schema/provider.definition.schema.v1.json
anshikg: I am following the pattern for insertionOrder and letting drift handle this. Such this is not specific to a language

src/test/java/software/amazon/cloudformation/resource/ValidatorTest.java
anshikg: may be DRIFT_METADAT_SCHEMA_PATH?

aygold92 (changes requested):
I'm pretty against these changes.  Like @PatMyron mentioned, we're hacking data into the schema to support a CloudFormation feature, and it's not even complete support -- it's a half baked solution that only solves for a couple use cases.  Drift still has plenty of other equality issues to worry about.

I think the real solution for drift is to take a baseline of the live state at the end of an operation and use that to compare against instead of the template. We will never get the template to match perfectly

ikben:
I have been thinking about these changes a bit, and I have a few questions/concerns:

> Some properties can be updated by the downstream service teams and these properties could have a prefix/postfix or both once returned from the Read Handler

I have no idea what this means in practice. From the perspective of a third-party resource provider, there is no such thing as a "downstream service team".

- If I'm writing a resource for our own products, there are no "service teams"
- If I'm writing a resource to plug a gap in coverage for AWS resources, the only thing I get to interact with is an API.

I also have concerns that relate a bit to the remark by @jotompki  
> are we planning on handling this sort of prefix additions handler side? if so would have to do for every language plugin

Mainly that I would expect that the output of the ReadHandler would always be a valid input for each property (assuming they're not readOnly).

The ignore case and null is empty list should fit that expectation. The suffix and prefix maybe do not. 

- If I add this to a property that I can also !GetAtt or !Ref, will the output shown in the drift detection be meaningfully different than what I would see in an output? Having any difference between those sounds like something that might lead to problems. The difference changing the actual string even more so.
- A particular future problem would be a feature that automates the import of existing resources. If there is meta-data present, should it remove the suffix/prefix before generating a template? Depending on the schema, this might even break the contract [see below]
- If the underlying API changes, I would expect the handler to deal with that, the way I read the suffix/prefix (and I might be reading it wrong, see my first point), that responsibility is instead pushed to the caller.
 

----
from https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-read

> A read handler MUST return a model representation that conforms to the shape of the resource schema

If you specify a regex or similar for a property, the returned model might not fit in the shape

PatMyron:
> > these properties could have a prefix/postfix or both once returned from the Read Handler
> 
> I have no idea what this means in practice

@ikben [`AWS::Route53::HostedZone.Name`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-route53-hostedzone.html#cfn-route53-hostedzone-name) is a good example (`.` suffix):

Screen Shot 2020-10-07 at 1 44 33 PM

README.md
```diff
@@ -139,6 +139,25 @@ Together with the `uniqueItems` property (which is native to JSON Schema), compl
 | true           | true           | ordered set    |
 | false          | true           | set      |
 
+#### optionalPrefix/optionalSuffix
+
+Some properties can be updated by the downstream service teams and these properties could have a prefix/suffix or both once returned from the Read Handler. These metadata properties are strings
+which can be provided for a property and checked on drift to ensure the property has not drifted.
+
+`optionalPrefix: "anyString"`
+`optionalSuffix: '"anyString"`
```
benkehoe, line 148: Extraneous single quote

jotompki:
> > > these properties could have a prefix/postfix or both once returned from the Read Handler
> > 
> > 
> > I have no idea what this means in practice
> 
> @ikben [`AWS::Route53::HostedZone.Name`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-route53-hostedzone.html#cfn-route53-hostedzone-name) is a good example (optional `.` suffix):
> 
> Screen Shot 2020-10-07 at 1 44 33 PM

Are there any other use cases other than this ? This could set a precedent of adding in metadata to the schema for one-off drift integration issues.

PatMyron:
> Are there any other use cases other than this ? This could set a precedent of adding in metadata to the schema for one-off drift integration issues.

I do worry if this problem is restricted to only prefixes and suffixes too. For example, many times `.` don't matter in the first half of email addresses, so this problem doesn't seem strictly limited to prefixes and suffixes

anshikg:
> > > > these properties could have a prefix/postfix or both once returned from the Read Handler
> > > 
> > > 
> > > I have no idea what this means in practice
> > 
> > 
> > @ikben [`AWS::Route53::HostedZone.Name`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-route53-hostedzone.html#cfn-route53-hostedzone-name) is a good example (optional `.` suffix):
> > Screen Shot 2020-10-07 at 1 44 33 PM
> 
> Are there any other use cases other than this ? This could set a precedent of adding in metadata to the schema for one-off drift integration issues.

IAM Policy document has values which expect an account Id but return Arn.

PatMyron:
> IAM Policy document has values which expect an account Id but return Arn.

Can these optional prefixes/suffixes be patterns instead of literal strings? This doesn't seem solvable by literal string prefixes/suffixes since [ARNs differ per partition](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax)

ikben:
>  AWS::Route53::HostedZone.Name is a good example (optional . suffix):

That is a good example as the behavior is well defined by the API. If this is the kind of use case, I think it would make sense to define it in the readme as strings that are 100% equivalent.

> IAM Policy document has values which expect an account Id but return Arn.

That sounds to me like a bug in the resource provider as it breaks the contract ([AWS::IAM::Policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-policy.html) defines a regex that wouldn't match an ARN for eg. Groups)

For cases where it wouldn't break the contract (eg. the Property can be either a name or an ARN), you would have to be careful about how you do the conversion, as the conversion isn't necessarily symmetric (if you remove `arn:aws:$service:.*:.*:thing/` from every input, things from different regions will collapse down to the same name)

wbingli:
To address the drift issues, there are different layers of solution. This is one of the solutions that to provide schema level metadata to address majority of drift issues without writing code.    I agree with the opinion https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/103#issuecomment-709625526 that the metadata name is confusing. Customer may think whether the property name is `caseSensitive` or not, and wont think it's for drift feature.   Something I like is following:

```
drift-override : {
 "caseSensitive": false,
 "nullEqualsEmpty": true
 }
```

Other layer of solution is providing more flexible and dynamic solution so that resource handler can make decision.  

As for Andrew's comment that "I think the real solution for drift is to take a baseline of the live state at the end of an operation and use that to compare against instead of the template."https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/103#pullrequestreview-510740117.   My main concern to this approach is that the live state of last operation is different from the CFN template. It will bring customer UX issue to map yet another model with CFN template to figure out what is changed or not.

anshikg:
Closing this pull request. This is not a scalable mechanism to support this feature.

PR #104, opened 2020-11-09
Adding support to fetch handler permissions from schema object

jotompki:
*Issue #, if available:*

*Description of changes:* Allows for access of handler permissions via the schema object. Can now retrieve permissions in the following way:

```
ResourceTypeSchema schema = ResourceTypeSchema.load(resourceDefinition);
//Set of permissions required to create a resource
Set<String> createPermissions = schema.getHandlerPermissions.get("create");
```

I left the handler properties as unprocessed as timeoutInMinutes is still yet to be processed. If necessary I can add in some logic to set another member variable to process these, but there's no use currently for fetching these.


PatMyron:
What's this for btw? Wondering what'll use this vs. parsing JSON directly

ammokhov (approved):

jotompki:
> What's this for btw? Wondering what'll use this vs. parsing JSON directly

Was working on some items that required comparing the permissions and this seemed to be a more sharable way to do the work. If something is modelled in the schema it should be retrievable in the schema object. Parsing the json directly works but wanted to provide an easy way to get them in case a similar thing needs to be done elsewhere.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -110,6 +113,16 @@ public ResourceTypeSchema(Schema schema) {
             ((ArrayList<?>) v).forEach(p -> this.writeOnlyProperties.add(new JSONPointer(p.toString())));
             return null;
         });
+
+        this.unprocessedProperties.computeIfPresent("handlers", (k, v) -> {
+            ((HashMap<?, ?>) v).keySet().forEach(handlerKey -> {
+                handlerPermissions.put(handlerKey.toString(), new HashSet<>());
+                List<?> permissionsList = (List<?>) ((HashMap<?, ?>) ((HashMap<?, ?>) v).get(handlerKey)).get("permissions");
```
wbingli, line 120: Can we cast it to `List<String>` for the permissions?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -110,6 +113,16 @@ public ResourceTypeSchema(Schema schema) {
...
+                permissionsList.forEach(permission -> handlerPermissions.get(handlerKey.toString()).add(permission.toString()));
+            });
+            // timeoutInMinutes still unprocessed
```
wbingli, line 123: Maybe a good way is to model the permissions + timeoutInImutes as a POJO such as "Handler".  Then we have separated method to retrieve them

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
```diff
@@ -244,4 +245,20 @@ static JSONObject getEmptyModel() {
...
+    /**
+     * validate that handler permissions are processed and can be retrieved programatically via the schema object
+     */
+
+    @Test
+    public void getHandlerPermissions_validPermissions_shouldReturnHandlerSetMap() {
+        JSONObject resourceDefinition = loadJSON(HANDLERS_SCHEMA_PATH);
+        ResourceTypeSchema schema = ResourceTypeSchema.load(resourceDefinition);
+
+        assertThat(schema.getHandlerPermissions().get("create")).contains("test:permission");
```
wbingli, line 257: I don't like the way to expose internal data as interface, espeically the Map. Others could protential modify it such as `schema.getHandlerPermissions().set("create", new HashSet<>())`.

I would suggest we should expose it as public method, such as 

```
public Set<String> getHandlerPermissions(final String handlerName) {
  return this.schema.getHandlerPermissions().get("create").getPermissions();
}

```

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
jotompki: agree with that. i can restructure so this is the case

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -148,6 +167,14 @@ public String getDescription() {
         return this.writeOnlyProperties.stream().map(JSONPointer::toString).collect(Collectors.toList());
     }
 
+    public Set<String> getHandlerPermissions(String action) {
```
wbingli, line 170: Return `null` is generally not a good practice as they need to check null for the result and it's easy to forget.

Let's provide other method ` boolean hasHandler(final String action)` so that they can use it to check if it has handler before to call those methods.

jotompki: cool i can do this. what should we do in cases where there isn't  a handler though? are you just saying to add the `hasHandler` and keep this the same?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -110,6 +116,19 @@ public ResourceTypeSchema(Schema schema) {
             ((ArrayList<?>) v).forEach(p -> this.writeOnlyProperties.add(new JSONPointer(p.toString())));
             return null;
         });
+
+        this.unprocessedProperties.computeIfPresent("handlers", (k, v) -> {
```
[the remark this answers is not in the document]
jotompki: my understanding of `computeIfPresent` leads me to believe if we are "processing" the `unprocessedProperties`, returning null removes the key from the map, marking it as processed. 

This leads me to believe this is what we are trying to do: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfPresent-K-java.util.function.BiFunction-

"If the function returns null, the mapping is removed. "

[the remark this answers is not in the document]
wbingli: I see. It's kind of processing and remove the property. If so, let's keep it this way.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
wbingli: I'm kind of on the fence to enforce something like using `Optional<Set<String>>`, it could be overkill.  The same practice is Java `Map` interface. The `get` return `null` if doesn't exist or the value could be null. But need to at least provide an existence check method.

Let's keep it this way and provide an extra `hasHandler` method.

wbingli (approved):

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -51,6 +55,8 @@
     private final List<List<JSONPointer>> additionalIdentifiers = new ArrayList<>();
     private final List<JSONPointer> readOnlyProperties = new ArrayList<>();
     private final List<JSONPointer> writeOnlyProperties = new ArrayList<>();
+    @Getter(AccessLevel.NONE)
```
wbingli, line 58: Let's remove the `@Getter` here

jotompki: this suppresses the getter so someone can't do something like:

```
Map<String, Handlers> handlerMap = schema.getHandlers();
handlerMap.put("create", new Handler());
```

Setting the access level to none overrides the @Getter at the top of the class:
https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/1e99ac138807111f16c5368cf6977ec09dbd6d42/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java#L35-L39

wbingli: Got it.

PR #105, opened 2020-12-10
Adding `taggable` support in schema

AugustXiaodm:
*Issue #, if available:*

Adding `taggable` support for resource schema. This boolean flag indicates whether a resource type supports updatable tagging property. 

*Description of changes:*
Added `taggable` support.
Updated Junit dependencies to the latest versions. (supporting booleans as test ValueSource)


README.md
```diff
@@ -36,6 +36,7 @@ Certain properties of a resource are _semantic_ and have special meaning when us
...
 * **`replacementStrategy`**: As mentioned above, changing a `createOnlyProperty` requires replacement of the resource by creating a new one and deleting the old one. The default CloudFormation replacement behavior is to create a new resource first, then delete the old resource, so as to avoid any downtime. However, some resources are singleton resources, meaning that only one can exist at a time. In this case, it is not possible to create a second resource first, so CloudFormation must Delete first and then Create. Specify either `create_then_delete` or `delete_then_create`. Default value is `create_then_delete`
+ * **`taggable`**: A boolean type property which defaults to true, indicating this resource type supports updatable [`tagging property`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html). Otherwise, it indicates this resource type does not contain any updatable [`tagging properties`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).
```
PatMyron, line 39: not sure I understand why `Tags` specifically needs this. [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) marks which properties are not updatable, and can already be applied to `Tags` property types

AidenAtSea (approved):

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -198,6 +204,10 @@ public String getReplacementStrategy() {
         return this.replacementStrategy;
     }
 
+    public boolean getTaggable() {
```
jotompki, line 207: doesn't lombok handle this? isTaggable()?

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
```diff
@@ -121,6 +124,7 @@ public void getSingletonReplacementStrategy() {
 
         String replacementStrategy = schema.getReplacementStrategy();
         assertThat(replacementStrategy).isEqualTo("delete_then_create");
+        assertTrue(schema.getTaggable());
```
jotompki, line 127: nit: 

can we use a pattern like assertThat().isTrue()

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
AugustXiaodm: Yes, lombok generated `isTaggable` for it. Thanks.

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
AugustXiaodm: Sure. Thanks.

wenyhu (approved):

AidenAtSea (approved):

README.md
aygold92: yeah its essentially meant to indicate "does this resource support tagging?" We can't just check for the presence of a "Tags" property because some resources call it different things like "EC2Tags" or something like that.

aygold92 (approved):

README.md
PatMyron: >We can't just check for the presence of a "Tags" property because some resources call it different things like "EC2Tags" or something like that.

@aygold92 do you have specific examples btw? Haven't been able to find any myself yet

AidenAtSea: One example
AWS::EFS::AccessPoint -> AccessPointTags

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-accesspoint.html#cfn-efs-accesspoint-accesspointtags

PatMyron:
[It'd be nice to schematize more tagging information like support for tag-on-create, system tags, etc](https://twitter.com/InesParnisari/status/1327040431138299904):
https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/228

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -301,6 +301,11 @@
             ],
             "$ref": "#/definitions/httpsUrl"
         },
+        "taggable": {
+            "$comment": "A boolean flag indicating whether this resource type supports updatable tagging.",
+            "type": "boolean",
+            "default": "true"
```
PatMyron, line 307: @Yunhao-Jiao had a good catch that we know `taggable` is `false` if an `UpdateHandler` isn't present
```suggestion
            "type": "boolean"
```
think it might make more sense to default to `true` if there's an `UpdateHandler`, and `false` if not

PatMyron: especially since only a few of the [existing 38 AWS resource providers without `UpdateHandler`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) set `taggable`:
```bash
CloudformationSchema $ grep -L '"update" : {' * | wc -l # resource schemas without "update"
      38
CloudformationSchema $ grep -L '"update" : {' * | xargs grep 'taggable'
aws-backup-backupselection.json:  "taggable" : false,
aws-route53-dnssec.json:  "taggable" : false,
aws-servicecatalogappregistry-attributegroupassociation.json:  "taggable" : false,
aws-servicecatalogappregistry-resourceassociation.json:  "taggable" : false,
aws-ses-configurationset.json:  "taggable" : false,
```

PatMyron: https://github.com/aws-cloudformation/cloudformation-resource-schema/pull/122

aygold92: the whole issue with lack of update handler is a CFN specific issue, and that is orthogonal to the issue here.  A resource can be taggable (i.e., it supports tags) and also not have an update handler.  Taggable is true by default because we expect that all resources should support tags

PatMyron: I honestly think `taggable` is too nuanced to be adequately captured by a `boolean`:
https://github.com/aws-cloudformation/cloudformation-resource-schema/pull/105#issuecomment-827975377
but assuming we stick with the definition introduced in this PR, which is that `taggable` means updatable tags, https://github.com/aws-cloudformation/cloudformation-resource-schema/pull/122 seems necessary to me

PR #106, opened 2020-12-15
Add property Transform in Uluru Schema to avoid falsely drifted resources

anshikg:
*Issue #, if available:* 
*Description of changes:* Avoid Falsely drifted resources by adding transformation function for every falsely drifted property


PatMyron (changes requested):
confused by `base.definition.schema.v1.json`, is that file coming from somewhere else? We only seem to be using a few lines of it, so I'm confused about the rest of it. I think I'd prefer just inlining the few lines we're using than splitting into multiple JSON schemas

PatMyron:
> confused by `base.definition.schema.v1.json`, is that file coming from somewhere else? We only seem to be using a few lines of it, so I'm confused about the rest of it. I think I'd prefer just inlining the few lines we're using than splitting into multiple JSON schemas

looks addressed in https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/107, closing this PR in favor of that one

PR #107, opened 2020-12-15
Add property Transform in Uluru Schema to avoid falsely drifted resources 

anshikg:
*Issue #, if available:*

*Description of changes:* Avoid Falsely drifted resources by adding transformation function for every falsely drifted property

### DIFF
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -422,6 +422,14 @@
         "resourceLink": {
             "description": "A template-able link to a resource instance. AWS-internal service links must be relative to the AWS console domain. External service links must be absolute, HTTPS URIs.",
             "$ref": "#/definitions/resourceLink"
+        },
+        "propertyTransform": {
+            "type": "object",
+            "patternProperties": {
+                "^[A-Za-z0-9]{1,64}$": {
+                    "type": "string"
+                }
+            }
         }
     },
     "required": [
--- a/src/test/resources/test-schema.json
+++ b/src/test/resources/test-schema.json
@@ -30,6 +30,10 @@
             }
         }
     },
+    "propertyTransform": {
+        "/properties/propertyA": "$join([$string(test), propertyA])",
+        "/properties/propertyB": "$count(propertyB) = 0 ? null : propertyB"
+    },
     "required": [
         "propertyB"
     ],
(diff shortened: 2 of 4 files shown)


src/test/resources/test-schema.json
```diff
@@ -30,6 +30,10 @@
             }
         }
     },
+    "propertyTransform": {
+        "/properties/propertyA": "$join([$string(test), propertyA])",
+        "/properties/propertyB": "$count(propertyB) = 0 ? null : propertyB"
```
benbridts, line 35: Out of curiosity:
- which language is this?
- Is that information documented somewhere?

jotompki: Agree with this let's try to document this a bit better in what it actually means. All other schema properties have additional documentation about how exactly they function.

anshikg: Out of curiosity:

which language is this?
> This is jsonta: https://jsonata.org/
Is that information documented somewhere?
> I am working on adding more documentation for this soon. Should be able to update this package's README and the CFN public documentation for the same soon.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -422,6 +422,14 @@
         "resourceLink": {
             "description": "A template-able link to a resource instance. AWS-internal service links must be relative to the AWS console domain. External service links must be absolute, HTTPS URIs.",
             "$ref": "#/definitions/resourceLink"
+        },
+        "propertyTransform": {
```
srujithpoondla03, line 426: Please add clear description on what this property is used for?

srujithpoondla03 (changes requested):

src/main/resources/schema/provider.definition.schema.v1.json
anshikg: This is part of the documentation and REAME which I am planning to send out next

srujithpoondla03: My comment is to add the description similar to other properties. For example `resourceLink` property has clear description on what it is used for : "A template-able link to a resource instance. AWS-internal service links must be relative to the AWS console domain. External service links must be absolute, HTTPS URIs."

ninavolkmuth (approved):

srujithpoondla03 (approved):

PR #111, opened 2021-01-11
Add ReadMe for propertyTransform

anshikg:
*Issue #, if available:*

*Description of changes:* Add REAME for propertyTransfrom


README.md
```diff
@@ -36,7 +36,8 @@ Certain properties of a resource are _semantic_ and have special meaning when us
...
- * **`taggable`**: A boolean type property which defaults to true, indicating this resource type supports updatable [`tagging property`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html). Otherwise, it indicates this resource type does not contain any updatable [`tagging properties`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).
+* **`taggable`**: A boolean type property which defaults to true, indicating this resource type supports updatable [`tagging property`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html). Otherwise, it indicates this resource type does not contain any updatable [`tagging properties`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).
+* **`propertyTransform`**: Is a map (Map<String, String>) with the key set to property name and value to set to jsonata transformation function (https://jsonata.org/). This property is used to avoid falsely drifted resources. If the handler transforms the input to the resource to an expected value a transform function can be defined for this property to avoid drift.
```
PatMyron, line 40: This library doesn't look heavily used outside of JavaScript, so I'm curious how this is implemented

jotompki, line 40: how are nested properties handled? is it a fully specified path?

jotompki, line 40:
```suggestion
* **`propertyTransform`**: Is a map (Map<String, String>) with the keys being property names and values being jsonata transformation functions (https://jsonata.org/). This property is used to avoid falsely drifted resources. If the handler transforms the input to the resource to an expected value a transform function can be defined for this property to avoid drift.
```

jotompki: might also be good to link to drift documentation as well if people are curious about what drift is

PatMyron: should have this format be consistent with the others

srujithpoondla03:
I think you have mentioned in the [previous PR](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/107) that you will add the [description](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/master/src/main/resources/schema/provider.definition.schema.v1.json#L426-L431) in meta-schema for `propertyTransform`. Please add the description.

README.md
anshikg: nested properties can be mentioned by appending dots. For example:
```
"foo": {
    "bar": {
       "type": "string"
    }
}

"propertyTransform": {
    "/properties/foo/bar": "$join([foo.bar, "test"])"
}
```

anshikg: Pat, I agree that we need more documentation. Right now the downstream service fix is still in progress and is not ready for customers to use. I am a little apprehensive about adding a whole lot of information without the deployment being complete. Please let me know if you disagree.

benbridts: Is there a good way to indicate an "in progress" property? From a customer perspective: having the documentation available before the implementation, seems like a good way to work backwards and provide feedback on how things might work. I completely understand the other side of that too (and I don't have a strong opinion either way).  

Currently, the only feedback that I have is very minor (although I'm sure more eyes would lead to more feedback), and something you'd probably address when creating more documentation:
- "If the handler transforms the input to the resource to an expected value a transform function can be defined for this property to avoid drift" was a bit hard to parse for me. If I read that correctly it could be rewritten as "If the read handler implementation is causing equivalent values to be incorrectly seen as drifted,  a `propertyTransform` can be applied to the (original) template properties, to make sure that the values being compared are using the same representation. This happens before the drift comparison is done"

README.md
```diff
@@ -36,7 +36,8 @@ Certain properties of a resource are _semantic_ and have special meaning when us
...
- * **`taggable`**: A boolean type property which defaults to true, indicating this resource type supports updatable [`tagging property`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html). Otherwise, it indicates this resource type does not contain any updatable [`tagging properties`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).
+* **`taggable`**: A boolean type property which defaults to true, indicating this resource type supports updatable [`tagging property`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html). Otherwise, it indicates this resource type does not contain any updatable [`tagging properties`](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).
+* **`propertyTransform`**: Is a map (Map<String, String>) with the keys being property names and values being jsonata transformation functions (https://jsonata.org/). This property is used to avoid falsely drifted resources. If the handler transforms the input to the resource to an expected value a transform function can be defined for this property to avoid drift.
```
jotompki, line 40:
```suggestion
* **`propertyTransform`**: Is a map (Map<String, String>) with the keys being property paths and values being jsonata transformation functions (https://jsonata.org/). This property is used to avoid falsely drifted resources. If the handler transforms the input to the resource to an expected value a transform function can be defined for this property to avoid drift.
```

jotompki (approved):

wbingli (approved):

Issue #30, opened 2019-09-07
Schematize nuanced immutability (conditional resource replacement, downtime, etc.)


rjlohan:
Currently, the CloudFormation resource specification and the documentation do not distinguish the difference between an update operation which _fully replaces_ a resource (including data loss) from those updates which cause some other kind of downtime (e.g a server or database reboot).

The documentation maps both of these concepts to 'Update requires: some interruption' and the resource spec maps only the fully immutable replacement operation, as does the schema currently.

We could include a schema property to signal that a property may cause downtime without data loss as a distinguishing concept.

PatMyron:
@rjlohan trying to assemble a list of likely stateful resource types here to help prevent data loss:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1232

PatMyron:
In addition to this problem with `createOnlyProperties` being more nuanced than being expressable in just a boolean, the same problem applies to `readOnlyProperties` and `writeOnlyProperties` as well

PR #120, opened 2021-03-24
Add compare method for create only properties

anshikg:
*Issue #, if available:*

*Description of changes:*
Right now, if the create only property passed to the update handler is different from the created model it is the handler's responsibility to throw a not update-able exception. This works for now. But in my opinion we should have this check in the language plugins and support this functionality on the framework vs asking the handlers to throw the right error code. This CR is creating a method to compare the previous model's create only property with the updated model's create only properties and if they are different the appropriate exception is going to be thrown by the language plugins. (Validation in language plugins will be a separate CR)

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -206,6 +206,18 @@ public class ResourceTypeSchema {
             new PublicJSONPointer(writeOnlyProperty.replaceFirst("^/properties", "")), resourceModel));
     }
 
+    public boolean compareCreateOnlyProperties(final JSONObject previousResourceModel,
+                                               final JSONObject currentResourceModel) {
+        boolean compare = true;
+        for (String createOnlyProperty : this.getCreateOnlyPropertiesAsStrings()) {
+            final boolean result = compareProperty(
+                new PublicJSONPointer(createOnlyProperty.replaceFirst("^/properties", "")),
+                previousResourceModel, currentResourceModel);
+            compare = compare && result;
+        }
+        return compare;
+    }
+
     public String getReplacementStrategy() {
         return this.replacementStrategy;
     }
@@ -226,6 +238,29 @@ public class ResourceTypeSchema {
             }
         } catch (JSONPointerException | NumberFormatException e) {
             // do nothing, as this indicates the model does not have a value for the pointer
+        }
+    }
+
+    public static boolean compareProperty(final PublicJSONPointer property,
+                                          final JSONObject previousResourceModel,
+                                          final JSONObject currentresourceModel) {
+        List<String> refTokens = property.getRefTokens();
+        final String key = refTokens.get(refTokens.size() - 1);
+        try {
+            // if size is more than one, fetch parent object/array of key to remove so that
+            // we can remove
+            if (refTokens.size() > 1) {
+                // use sublist to specify to point at the parent object
+                final JSONPointer parentObjectPointer = new JSONPointer(refTokens.subList(0, refTokens.size() - 1));
+                final JSONObject previousParentObject = (JSONObject) parentObjectPointer.queryFrom(previousResourceModel);
+                final JSONObject currentParentObject = (JSONObject) parentObjectPointer.queryFrom(currentresourceModel);
+                return previousParentObject.get(key) == currentParentObject.get(key);
+            } else {
+                return previousResourceModel.get(key) == currentresourceModel.get(key);
+            }
+        } catch (JSONPointerException | NumberFormatException e) {
+            // do nothing, as this indicates the model does not have a value for the pointer
+            return false;
         }
     }
 
--- a/src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
+++ b/src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
@@ -37,6 +37,8 @@ public class ResourceTypeSchemaTest {
     private static final String MINIMAL_SCHEMA_PATH = "/minimal-schema.json";
     private static final String NO_ADDITIONAL_PROPERTIES_SCHEMA_PATH = "/no-additional-properties-schema.json";
     private static final String WRITEONLY_MODEL_PATH = "/write-only-model.json";
+    private static final String CREATEONLY_MODEL_PATH = "/create-only-model.json";
+    private static final String CREATEONLY_CURRENT_MODEL_PATH = "/create-only-current-model.json";
     private static final String HANDLERS_SCHEMA_PATH = "/valid-with-handlers-schema.json";
 
     @Test
@@ -58,7 +60,8 @@ public class ResourceTypeSchemaTest {
 
         List<String> result = schema.getCreateOnlyPropertiesAsStrings();
 
-        assertThat(result).containsExactly("/properties/propertyA", "/properties/propertyD");
+        assertThat(result).containsExactly("/properties/propertyA", "/properties/propertyD",
+            "/properties/propertyE/nestedCreateOnlyProperty");
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


PatMyron:
1) @aygold92 wondering if platform can generally handle this instead of handling this in each plugin / in each handler

2) @ammokhov this also needs to consider [`conditionallyCreateOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30) once we have those modelable

aygold92:
can you be more specific in the description (and inline too, probably) what this is "comparing"/for what purpose?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -206,6 +206,18 @@ public void removeWriteOnlyProperties(final JSONObject resourceModel) {
             new PublicJSONPointer(writeOnlyProperty.replaceFirst("^/properties", "")), resourceModel));
     }
 
+    public boolean compareCreateOnlyProperties(final JSONObject previousResourceModel,
+                                               final JSONObject currentResourceModel) {
+        boolean compare = true;
+        for (String createOnlyProperty : this.getCreateOnlyPropertiesAsStrings()) {
+            final boolean result = compareProperty(
```
jotompki, line 213: if you hit a failure just fail out instead of checking more properties

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -206,6 +206,18 @@ public void removeWriteOnlyProperties(final JSONObject resourceModel) {
             new PublicJSONPointer(writeOnlyProperty.replaceFirst("^/properties", "")), resourceModel));
     }
 
+    public boolean compareCreateOnlyProperties(final JSONObject previousResourceModel,
```
jotompki, line 209: should this go into the schema class? seems like it more just utilizes the schema's properties. are we using this in multiple places? if not can we just include the logic where  we are doing the comparisons?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -229,6 +241,29 @@ public static void removeProperty(final PublicJSONPointer property, final JSONOb
...
     }
 
+    public static boolean compareProperty(final PublicJSONPointer property,
+                                          final JSONObject previousResourceModel,
+                                          final JSONObject currentresourceModel) {
+        List<String> refTokens = property.getRefTokens();
+        final String key = refTokens.get(refTokens.size() - 1);
+        try {
+            // if size is more than one, fetch parent object/array of key to remove so that
+            // we can remove
```
jotompki, line 251: comment not accurate

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -229,6 +241,29 @@ public static void removeProperty(final PublicJSONPointer property, final JSONOb
         }
     }
 
+    public static boolean compareProperty(final PublicJSONPointer property,
+                                          final JSONObject previousResourceModel,
+                                          final JSONObject currentresourceModel) {
```
jotompki, line 246:
```suggestion
                                          final JSONObject currentResourceModel) {
```

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -229,6 +241,29 @@ public static void removeProperty(final PublicJSONPointer property, final JSONOb
...
+                // use sublist to specify to point at the parent object
+                final JSONPointer parentObjectPointer = new JSONPointer(refTokens.subList(0, refTokens.size() - 1));
+                final JSONObject previousParentObject = (JSONObject) parentObjectPointer.queryFrom(previousResourceModel);
+                final JSONObject currentParentObject = (JSONObject) parentObjectPointer.queryFrom(currentresourceModel);
+                return previousParentObject.get(key) == currentParentObject.get(key);
+            } else {
+                return previousResourceModel.get(key) == currentresourceModel.get(key);
+            }
+        } catch (JSONPointerException | NumberFormatException e) {
+            // do nothing, as this indicates the model does not have a value for the pointer
```
jotompki, line 262: if both don't have a property would it trip this case?

jotompki (changes requested):
What's this to be used for?

ammokhov (changes requested):
>     1. @aygold92 wondering if platform can generally handle this instead of handling this in each plugin / in each handler
> 
>     2. @ammokhov this also needs to consider [`conditionallyCreateOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30) once we have those modelable

good point. i feel like if we add it to java plugin then it becomes a hard requirement on any other future plugins, which is not obvious. If we go that route then we should have a strict requirements on what plugin must satisfy, which we currently dont have. 

I also think this is a bit redundant and probably not necessary

ammokhov (changes requested):
additionally, I dont think this is the best place to put immutability analyzer. 

this is invoked twice in java plugin: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L201
which is to hide deserialization mismatch

and the other one: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L291
when we already published metrics and started request transformation

anshikg:
> What's this to be used for?

Updated the details section with the usage

anshikg:
> > ```
> > 1. @aygold92 wondering if platform can generally handle this instead of handling this in each plugin / in each handler
> > 
> > 2. @ammokhov this also needs to consider [`conditionallyCreateOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30) once we have those modelable
> > ```
> 
> good point. i feel like if we add it to java plugin then it becomes a hard requirement on any other future plugins, which is not obvious. If we go that route then we should have a strict requirements on what plugin must satisfy, which we currently dont have.
> 
> I also think this is a bit redundant and probably not necessary

This is already a hard requirement as this is asserted by Contract test: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update_invalid.py#L20
All this CR is doing is pushing the validation upstream in the framework vs having the check in the handlers. Also, this is also one of the handler contracts:
``` Any createOnlyProperties specified in update handler input MUST NOT be different from their previous state.```

anshikg:
> additionally, I dont think this is the best place to put immutability analyzer.
> 
> this is invoked twice in java plugin: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L201
> which is to hide deserialization mismatch
> 
> and the other one: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L291
> when we already published metrics and started request transformation

Could you help me understand why ``` I dont think this is the best place to put immutability analyzer.``` ?
If we are making the code change for all plugins I think have this code in a central location helps avoid code replication

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
anshikg: thats a good catch, we should not be failing in this condition

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
anshikg: We will be using this for all plugins. This was the reason I added it to schema package

PatMyron (changes requested):
> Also, this is also one of the handler contracts:
> ` Any createOnlyProperties specified in update handler input MUST NOT be different from their previous state.`

This falls under ["The update handler can make the following assumptions about input submitted to it"](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-update) which means individual UpdateHandlers _shouldn't_ have to worry about handling this

> This is already a hard requirement as this is asserted by Contract test

As mentioned in https://github.com/aws-cloudformation/cloudformation-cli/issues/527#issuecomment-680696823, it _shouldn't_ be

ammokhov:
> > additionally, I dont think this is the best place to put immutability analyzer.
> > this is invoked twice in java plugin: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L201
> > which is to hide deserialization mismatch
> > and the other one: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/AbstractWrapper.java#L291
> > when we already published metrics and started request transformation
> 
> Could you help me understand why ` I dont think this is the best place to put immutability analyzer.` ?
> If we are making the code change for all plugins I think have this code in a central location helps avoid code replication

this project is published in maven which is compatible only for java plugin. other plugins do not use schema project whatsoever. python plugin using different package manager, so as go. This change equates as if you made it inside java plugin

ammokhov:
> > > ```
> > > 1. @aygold92 wondering if platform can generally handle this instead of handling this in each plugin / in each handler
> > > 
> > > 2. @ammokhov this also needs to consider [`conditionallyCreateOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30) once we have those modelable
> > > ```
> > 
> > 
> > good point. i feel like if we add it to java plugin then it becomes a hard requirement on any other future plugins, which is not obvious. If we go that route then we should have a strict requirements on what plugin must satisfy, which we currently dont have.
> > I also think this is a bit redundant and probably not necessary
> 
> This is already a hard requirement as this is asserted by Contract test: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update_invalid.py#L20
> All this CR is doing is pushing the validation upstream in the framework vs having the check in the handlers. Also, this is also one of the handler contracts:
> ` Any createOnlyProperties specified in update handler input MUST NOT be different from their previous state.`

contract tests are behavioral asserts for handlers. not plugins. plugin is a facade layer that all handlers share and we dont have any official requirements for them as of now. By adding this change we imposing a MUST implement feature on every plugin. 

If we want to remove this burden from handlers then it should be done on the real upstream level - all the services that invoke handlers but not the plugin which is pretty much the same level as handlers

anshikg:
Checked with stakeholders offline, This CR is redundant and these checks are being made upstream. Therefore closing this CR

PR #121, opened 2021-03-26
Adding conditionallyCreateOnlyProperties to metaschema

ammokhov:
*Issue #, if available:* https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30

*Description of changes:*

Adding `conditionallyCreateOnlyProperties` to meta schema, so that resource providers could model properties that could not deterministically predict replacement/inline update

### DIFF
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -387,8 +387,8 @@
             "description": "A list of JSON pointers to properties (typically sensitive) that are able to be specified by the customer but unable to be returned in a Read request",
             "$ref": "#/definitions/jsonPointerArray"
         },
-        "conditionallyCreateOnlyProperties": {
-            "description": "A list of JSON pointers to properties that are only able to cause replacement during the Update request.",
+        "conditionalCreateOnlyProperties": {
+            "description": "A list of JSON pointers for properties that can only be updated under certain conditions. For example, you can upgrade the engine version of an RDS DBInstance but you cannot downgrade it.  When updating this property for a resource in a CloudFormation stack, the resource will be replaced if it cannot be updated.",
             "$ref": "#/definitions/jsonPointerArray"
         },
         "createOnlyProperties": {
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -387,6 +387,10 @@
             "description": "A list of JSON pointers to properties (typically sensitive) that are able to be specified by the customer but unable to be returned in a Read request",
             "$ref": "#/definitions/jsonPointerArray"
         },
+        "conditionallyCreateOnlyProperties": {
+            "description": "A list of JSON pointers to properties that are only able to cause replacement during the Update request.",
+            "$ref": "#/definitions/jsonPointerArray"
+        },
         "createOnlyProperties": {
             "description": "A list of JSON pointers to properties that are only able to be specified by the customer when creating a resource. Conversely, any property *not* in this list can be applied to an Update request.",
             "$ref": "#/definitions/jsonPointerArray"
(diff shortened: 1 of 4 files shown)


src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -387,6 +387,10 @@
             "description": "A list of JSON pointers to properties (typically sensitive) that are able to be specified by the customer but unable to be returned in a Read request",
             "$ref": "#/definitions/jsonPointerArray"
         },
+        "conditionallyCreateOnlyProperties": {
+            "description": "A list of JSON pointers to properties that are only able to cause replacement during the Update request.",
```
PatMyron, line 391: think `only` might be misleading. Also, do you think it's worth adding an example? (need to scroll to see suggestion)
```suggestion
            "description": "A list of JSON pointers to properties that may or may not cause replacement during the Update request. For example, a version property might be upgradable in-place but downgrading may require resource replacement",
```

ammokhov: You think we should have a specific example like version?

anshikg (approved):

jotompki (approved):
Looks good pending the change to the description

src/main/resources/schema/provider.definition.schema.v1.json
ammokhov: @jotompki @wbingli @aygold92 @anshikg thoughts?

aygold92: I would even give a specific example, like that for EBS volumes you can increase the `size` property, but decreasing it would require replacement https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/modify-volume-requirements.html#elastic-volumes-limitations

That being said, I would not even mention replacement, or if so mention that specifically as a cloudformation specific thing.  Cloud API will never replace your resource, it would just fail with `NotUpdatable` exception.  I would say something like:

```
A list of JSON pointers for properties that can only be updated under certain conditions.  For example, you can increase the size of an EBS volume but you cannot decrease it.  When updating this property for a resource in a CloudFormation stack, the resource will be replaced if it cannot be updated.
```

PatMyron: [One consideration about specific examples is that they may change in the future](https://github.com/awsdocs/aws-cloudformation-user-guide/pull/904). EBS might remove the restriction on decreasing `size` in-place eventually

aygold92:
> Could I suggest these be called `conditionalCreateOnlyProperties`? Adverbs in field names are pretty uncommon, and as a user I think I would often forget that it was `conditionallyCreateOnlyProperties`. It's a create only property that is conditional, rather than a property that is conditionally create only

I at first thought to call it `conditionalCreateOnlyProperties`, but I'm not sure what the distinction is between "a create only property that is conditional" and "a property that is conditionally create only".  Actually, it seems more like the second one in this case

src/main/resources/schema/provider.definition.schema.v1.json
aygold92: true, I'm fine with the generic example you had above

benkehoe:
My main point is that it seems unusually phrased, and thus easy to forget/get confused. But in my mind, it's a create-only property, that isn't always triggered, as opposed to be only sometimes create-only.

src/main/resources/schema/provider.definition.schema.v1.json
ammokhov: how about using rds? db downgrade is unlikely to change over time:
```
A list of JSON pointers for properties that can only be updated under certain conditions.  For example, you can upgrade the engine version of a RDS DBInstance but you cannot downgrade it.  When updating this property for a resource in a CloudFormation stack, the resource will be replaced if it cannot be updated.
```

benkehoe:
Introduced this feature to my TAM yesterday, and as I expected would happen, they echoed back the name as `conditionalCreateOnlyProperties`, so the confusion is already starting.

PR #122, opened 2021-05-20
Set default taggable to false when UpdateHandler not provided

Yunhao-Jiao:
*Issue #, if available:*
https://github.com/aws-cloudformation/cloudformation-resource-schema/pull/105#discussion_r634834873
*Description of changes:*
As discussed with @PatMyron offline, we need to set the default value of taggable to be false when UpdateHandler is not provided, so that update on stack level tags won't trigger the replace update of the resource.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -92,11 +92,6 @@ public class ResourceTypeSchema {
             : "create_then_delete";
         this.unprocessedProperties.remove("replacementStrategy");
 
-        this.taggable = this.unprocessedProperties.containsKey("taggable")
-            ? Boolean.valueOf(this.unprocessedProperties.get("taggable").toString())
-            : true;
-        this.unprocessedProperties.remove("taggable");
-
         this.unprocessedProperties.computeIfPresent("conditionalCreateOnlyProperties", (k, v) -> {
             ((ArrayList<?>) v).forEach(p -> this.conditionalCreateOnlyProperties.add(new JSONPointer(p.toString())));
             return null;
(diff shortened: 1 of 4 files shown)


src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -304,7 +304,7 @@
         "taggable": {
             "$comment": "A boolean flag indicating whether this resource type supports updatable tagging.",
             "type": "boolean",
-            "default": true
+            "default": "Default to true if there's an UpdateHandler, and false if no UpdateHandler is provided."
```
PatMyron, line 307: maybe just remove `default` since I think it's for literal values, but this would be a good `description` to add
https://json-schema.org/understanding-json-schema/reference/generic.html

PatMyron (approved):

src/main/resources/schema/provider.definition.schema.v1.json
Yunhao-Jiao: Sure nice catch

PatMyron (approved):

wbingli (approved):

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +148,11 @@ public ResourceTypeSchema(Schema schema) {
             });
             return null;
         });
+
+        this.taggable = this.unprocessedProperties.containsKey("taggable")
```
wbingli, line 152: This may only be useful if the client will use `ResourceTypeSchema` and reply on this package.

aygold92:
I think we should revert this change, as I commented on #105 

```
the whole issue with lack of update handler is a CFN specific issue, and that is orthogonal to the issue here. A resource can be taggable (i.e., it supports tags) and also not have an update handler. Taggable is true by default because we expect that all resources should support tags
```

PR #128, opened 2021-07-06
Check if properties list is empty before attempting to remove

narmadn:
*Issue #, if available:*

This change add a check if property list is empty or not before it attempt to remove any property.


omkhegde (approved):

src/test/resources/empty-write-only-model.json
```diff
@@ -0,0 +1,5 @@
+{
+    "propertyE": {
+        "writeOnlyArray": []
+    }
+}
```
PatMyron, line 5: confused by this test case since it doesn't have any `writeOnlyProperties`

narmadn: Added additional schema input file just to avoid confusion and used the failed scenario as test case.

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
```diff
@@ -210,6 +212,18 @@ public void removeWriteOnlyProperties_hasWriteOnlyProperties_shouldRemove() {
         assertThat(resourceModel.has("propertyB")).isTrue();
     }
 
+    @Test
+    public void removeWriteOnlyProperties_hasWriteOnlyProperties_shouldDoNothing() {
+        JSONObject o = loadJSON(TEST_SCHEMA_WITH_EMPTY_WRITE_ONLY_PATH);
+        ResourceTypeSchema schema = ResourceTypeSchema.load(o);
+        JSONObject resourceModel = loadJSON(EMPTY_WRITEONLY_MODEL_PATH);
+
+        schema.removeWriteOnlyProperties(resourceModel);
```
PatMyron, line 221: still confused by the testing, doesn't make sense to me to call `removeWriteOnlyProperties` with another `resourceModel`, especially since that one doesn't have [`writeOnlyProperties`](https://github.com/aws-cloudformation/cloudformation-resource-schema#resource-semantics)
```suggestion
        JSONObject o = loadJSON(TEST_SCHEMA_WITH_EMPTY_WRITE_ONLY_PATH);
        ResourceTypeSchema schema = ResourceTypeSchema.load(o);

        schema.removeWriteOnlyProperties(o);
```

sungkkim (approved):

src/test/java/software/amazon/cloudformation/resource/ResourceTypeSchemaTest.java
narmadn: Built resourcemodel from same schema.

PR #131, opened 2021-09-01
Add tagging metadata into resource schema

Yunhao-Jiao:
*Issue #, if available:*

*Description of changes:*
Add tagging metadata into resource schema so that customers will know more about the tagging feature support status.

### DIFF
--- a/#/hooks/commit-msg
+++ b/#/hooks/commit-msg
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+git secrets --commit_msg_hook -- "$@"
--- a/#/hooks/commit-msg
+++ b/#/hooks/commit-msg
@@ -1,2 +1,0 @@
-#!/usr/bin/env bash
-git secrets --commit_msg_hook -- "$@"
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
@@ -25,6 +25,14 @@ import software.amazon.cloudformation.resource.exceptions.ValidationException;
 @Data
 @AllArgsConstructor
 public class ResourceTagging {
+    public static final String TAGGABLE = "taggable";
+    public static final String TAG_ON_CREATE = "tagOnCreate";
+    public static final String TAG_UPDATABLE = "tagUpdatable";
+    public static final String CLOUDFORMATION_SYSTEM_TAGS = "cloudFormationSystemTags";
+    public static final String TAG_PROPERTY = "tagProperty";
+    public static final ResourceTagging DEFAULT = new ResourceTagging(true, true, true,
+                                                                      true, new JSONPointer("/properties/Tags"));
+
     private boolean taggable;
     private boolean tagOnCreate;
     private boolean tagUpdatable;
@@ -32,19 +40,19 @@ public class ResourceTagging {
     private JSONPointer tagProperty;
 
     public void resetTaggable(final boolean taggableValue) {
-        taggable = taggableValue;
-        tagOnCreate = taggableValue;
-        tagUpdatable = taggableValue;
-        cloudFormationSystemTags = taggableValue;
+        this.taggable = taggableValue;
+        this.tagOnCreate = taggableValue;
+        this.tagUpdatable = taggableValue;
+        this.cloudFormationSystemTags = taggableValue;
     }
 
     public void validateTaggingMetadata(final boolean containUpdateHandler, final Schema schema) {
-        if (tagUpdatable && !containUpdateHandler) {
+        if (this.tagUpdatable && !containUpdateHandler) {
             throw new ValidationException("Invalid tagUpdatable value since update handler is missing", "tagging",
                                           "#/tagging/tagUpdatable");
         }
-        final String propertyName = tagProperty.toString().substring(tagProperty.toString().lastIndexOf('/') + 1);
-        if (taggable && !schema.definesProperty(propertyName)) {
+        final String propertyName = this.tagProperty.toString().substring(this.tagProperty.toString().lastIndexOf('/') + 1);
+        if (this.taggable && !schema.definesProperty(propertyName)) {
             final String errorMessage = String.format("Invalid tagProperty value since %s not found in schema", propertyName);
             throw new ValidationException(errorMessage, "tagging", "#/tagging/tagProperty");
         }
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
@@ -1,0 +1,52 @@
+/*
+* Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+package software.amazon.cloudformation.resource;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import org.everit.json.schema.JSONPointer;
+import org.everit.json.schema.Schema;
+
+import software.amazon.cloudformation.resource.exceptions.ValidationException;
+
+@Data
+@AllArgsConstructor
+public class ResourceTagging {
+    private boolean taggable;
+    private boolean tagOnCreate;
+    private boolean tagUpdatable;
+    private boolean cloudFormationSystemTags;
+    private JSONPointer tagProperty;
+
+    public void resetTaggable(final boolean taggableValue) {
+        taggable = taggableValue;
+        tagOnCreate = taggableValue;
+        tagUpdatable = taggableValue;
+        cloudFormationSystemTags = taggableValue;
+    }
+
+    public void validateTaggingMetadata(final boolean containUpdateHandler, final Schema schema) {
+        if (tagUpdatable && !containUpdateHandler) {
+            throw new ValidationException("Invalid tagUpdatable value since update handler is missing", "tagging",
+                                          "#/tagging/tagUpdatable");
+        }
+        final String propertyName = tagProperty.toString().substring(tagProperty.toString().lastIndexOf('/') + 1);
+        if (taggable && !schema.definesProperty(propertyName)) {
+            final String errorMessage = String.format("Invalid tagProperty value since %s not found in schema", propertyName);
+            throw new ValidationException(errorMessage, "tagging", "#/tagging/tagProperty");
+        }
+    }
+}
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -51,6 +51,7 @@ public class ResourceTypeSchema {
 
     private final String replacementStrategy;
     private final boolean taggable;
+    private final Map<String, Object> tagging = new HashMap<>();
     private final List<JSONPointer> createOnlyProperties = new ArrayList<>();
     private final List<JSONPointer> conditionalCreateOnlyProperties = new ArrayList<>();
     private final List<JSONPointer> deprecatedProperties = new ArrayList<>();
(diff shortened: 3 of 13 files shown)


src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +149,57 @@ public ResourceTypeSchema(Schema schema) {
             });
             return null;
         });
+
+        this.unprocessedProperties.computeIfPresent("tagging", (k, v) -> {
+            ((Map<?, ?>) v).forEach((key, value) -> {
+                if (!key.equals("tagProperty") && !key.equals("tagPropertyType")) {
```
wbingli, line 155: I would suggest you only do that for defined the property name.  It could cause issue or have unexcepted result if there are unknown properties.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +149,57 @@ public ResourceTypeSchema(Schema schema) {
...
+                    this.tagging.put(key.toString(), value.toString());
+                }
+            });
+
+            if (this.tagging.containsKey("taggable") && this.tagging.containsKey("tagOnCreate")
+                && this.tagging.get("taggable").equals(false) && this.tagging.get("tagOnCreate").equals(true)) {
+                throw new ValidationException("Invalid tagOnCreate value since taggable is marked false", "tagging",
+                                              "#/tagging/tagOnCreate");
+            } else if (this.tagging.containsKey("taggable") && this.tagging.containsKey("tagUpdatable") &&
+                this.tagging.get("taggable").equals(false) && this.tagging.get("tagUpdatable").equals(true)) {
```
wbingli, line 169: Can those 4 conditions be simplified with `getOrDefault` instead of `containsKey + get`?

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +149,57 @@ public ResourceTypeSchema(Schema schema) {
...
+                throw new ValidationException("Invalid tagUpdatable value since taggable is marked false", "tagging",
+                                              "#/tagging/tagUpdatable");
+            } else if (this.tagging.containsKey("taggable") && this.tagging.containsKey("CFNSystemTags") &&
```
wbingli, line 172: As those are validation check,  I would suggest to move into a separated method, and do not need `else if` for each validation.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +149,57 @@ public ResourceTypeSchema(Schema schema) {
...
+            if (!this.tagging.containsKey("taggable")) {
+                this.tagging.put("taggable", this.taggable);
+                if (this.taggable) {
+                    // default value to true when taggable to enforce tagging feature
+                    tagging.put("tagOnCreate", true);
+                    tagging.put("tagUpdatable", true);
+                    tagging.put("CFNSystemTags", true);
+                }
+            } else {
+                if (!this.tagging.get("taggable").equals(this.taggable)) {
```
wbingli, line 195: I would suggest we should have stricter rule. Only one `taggable` should be allowed to use. Even they are same value, we should enforce them to configure one.  In this way, we are deprecating the previous `taggable` config.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -114,10 +114,48 @@
...
+                    "description": "A boolean flag indicating whether this resource type supports tagging resources upon creation.",
+                    "type": "boolean",
+                    "default": true
+                },
+                "tagUpdatable": {
+                    "description": "A boolean flag indicating whether this resource type supports updatable tagging.",
+                    "type": "boolean",
+                    "default": true
+                },
+                "CFNSystemTags": {
```
wbingli, line 139: `CFNSystemTags`?  Don't use short abbr. for public name.  may `cloudFormationSystemTags`?

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -114,10 +114,48 @@
...
+                    "description": "A boolean flag indicating whether this resource type supports CloudFormation system tags.",
+                    "type": "boolean",
+                    "default": true
+                },
+                "tagProperty": {
+                    "description": "A reference to the Tags property in the schema.",
+                    "$ref": "http://json-schema.org/draft-07/schema#/properties/$ref"
+                },
+                "tagPropertyType": {
```
wbingli, line 148: If we know the `tagProperty`, we should be able to know the type by the reference?

#/hooks/commit-msg
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+git secrets --commit_msg_hook -- "$@"
```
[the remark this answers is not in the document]
Yunhao-Jiao: i see, will remove them

src/main/resources/schema/provider.definition.schema.v1.json
Yunhao-Jiao: sure, will change it

src/main/resources/schema/provider.definition.schema.v1.json
Yunhao-Jiao: yeah, will remove this attribute

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: that make sense, will add the validation

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: that make sense, changed it

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: yeah have changed this

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: sure have revised

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -51,6 +51,7 @@
 
     private final String replacementStrategy;
     private final boolean taggable;
+    private final Map<String, Object> tagging = new HashMap<>();
```
wbingli, line 54: We should model the `tagging` as a new class, e.g. `ResourceTagging`  instead of `Map<String, Object>`, it's hard to know what inside the Map.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +149,63 @@ public ResourceTypeSchema(Schema schema) {
...
+                throw new ValidationException("More than one configuration found for taggable value." +
+                    " Please remove the deprecated taggable property.", "tagging", "#/tagging/taggable");
+            }
+        } else {
+            this.taggable = true;
+        }
+        this.unprocessedProperties.remove("taggable");
+    }
+
+    private void validateTaggingMetadata() {
```
wbingli, line 188: If we have a new class, the validation could be part of that class.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -114,10 +114,43 @@
             "$ref": "base.definition.schema.v1.json#/definitions/httpsUrl"
         },
         "taggable": {
-            "description": "A boolean flag indicating whether this resource type supports tagging.",
+            "description": "(deprecated) A boolean flag indicating whether this resource type supports tagging.",
```
wbingli, line 117: As it's deprecated, we should suggest the right metadata to use.

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -114,10 +114,43 @@
...
+                    "default": true
+                },
+                "cloudFormationSystemTags": {
+                    "description": "A boolean flag indicating whether this resource type supports CloudFormation system tags.",
+                    "type": "boolean",
+                    "default": true
+                },
+                "tagProperty": {
+                    "description": "A reference to the Tags property in the schema.",
+                    "$ref": "http://json-schema.org/draft-07/schema#/properties/$ref"
```
wbingli, line 146: should the default value is `/properties/Tags`? we can validate it as well to ensure there is `/properties/Tags` otherwise, they have to configure the right  configure of it.

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: sounds good, yeah i can revise it

src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -114,10 +114,43 @@
...
         "taggable": {
-            "description": "A boolean flag indicating whether this resource type supports tagging.",
+            "description": "(deprecated) A boolean flag indicating whether this resource type supports tagging.",
             "type": "boolean",
             "default": true
         },
+        "tagging": {
+            "type": "object",
+            "properties": {
+                "taggable": {
```
sungkkim, line 124: Do we need `taggable`? Or can we assume it's true if the top level `tagging` property is supplied. Also, I think it would make more sense for it to be called something like tagSettings or tagConfig

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: yeah already moved it into the ResourceTagging class

src/main/resources/schema/provider.definition.schema.v1.json
Yunhao-Jiao: yeah we may still need taggable since some resources need to mark it false to skip related contract tests in the future.

Yunhao-Jiao: we have thought about tagConfig before as well. But even when service team config e.g. "tagOnCreate" to be true, they still need to implement this feature in handlers. We won't help them do implementation based on those contents, so they are more like simply metadata reflecting what are supported for now.

src/main/resources/schema/provider.definition.schema.v1.json
Yunhao-Jiao: sure already changed this and added validation

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -51,6 +51,9 @@
 
     private final String replacementStrategy;
     private final boolean taggable;
+    private final ResourceTagging tagging = new ResourceTagging(true, true, true,
```
wbingli, line 54: Could we have a default ResourceTagging, so we can do `ResourceTagging.DEFAULT`

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -153,6 +151,45 @@ public ResourceTypeSchema(Schema schema) {
             });
             return null;
         });
+
+        this.unprocessedProperties.computeIfPresent("tagging", (k, v) -> {
+            hasConfiguredTagging = true;
+            ((Map<?, ?>) v).forEach((key, value) -> {
+                if (key.equals("taggable")) {
```
wbingli, line 158: nit: Could we have the string as constant.  Those constant values could be defined in the class `ResourceTagging`, which should has all the knowledge of the resource tagging.

src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
```diff
@@ -0,0 +1,52 @@
...
+@AllArgsConstructor
+public class ResourceTagging {
+    private boolean taggable;
+    private boolean tagOnCreate;
+    private boolean tagUpdatable;
+    private boolean cloudFormationSystemTags;
+    private JSONPointer tagProperty;
+
+    public void resetTaggable(final boolean taggableValue) {
+        taggable = taggableValue;
```
wbingli, line 35: Could be better to `taggable -> this.taggable`

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: sure, let me add it

src/main/java/software/amazon/cloudformation/resource/ResourceTagging.java
Yunhao-Jiao: sure changed this

src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
Yunhao-Jiao: yeah added those fields

wbingli (approved):

zjinmei (approved):

PR #133, opened 2021-10-05
Add nonPublicProperties and nonPublicDefinitions lists of JSONPointers

WaelA:
*Description of changes:*

Add nonPublicProperties and nonPublicDefinitions lists of JSONPointers in metaschema. Adding top level lists to allow hidden
property and definition paths to be accepted. Also update lombok version as it fails local build.

### DIFF
--- a/src/main/resources/schema/provider.definition.schema.v1.json
+++ b/src/main/resources/schema/provider.definition.schema.v1.json
@@ -204,6 +204,14 @@
             "description": "A list of JSON pointers for properties that can only be updated under certain conditions. For example, you can upgrade the engine version of an RDS DBInstance but you cannot downgrade it.  When updating this property for a resource in a CloudFormation stack, the resource will be replaced if it cannot be updated.",
             "$ref": "base.definition.schema.v1.json#/definitions/jsonPointerArray"
         },
+        "nonPublicProperties": {
+            "description": "A list of JSON pointers for properties that are hidden. These properties will still be used but will not be visible",
+            "$ref": "base.definition.schema.v1.json#/definitions/jsonPointerArray"
+        },
+        "nonPublicDefinitions": {
+            "description": "A list of JSON pointers for definitions that are hidden. These definitions will still be used but will not be visible",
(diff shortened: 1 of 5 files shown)


src/main/resources/schema/provider.definition.schema.v1.json
```diff
@@ -204,6 +204,14 @@
             "description": "A list of JSON pointers for properties that can only be updated under certain conditions. For example, you can upgrade the engine version of an RDS DBInstance but you cannot downgrade it.  When updating this property for a resource in a CloudFormation stack, the resource will be replaced if it cannot be updated.",
             "$ref": "base.definition.schema.v1.json#/definitions/jsonPointerArray"
         },
+        "nonPublicProperties": {
+            "description": "A list of JSON pointers for properties that are hidden. These properties will still be used but will not be visible",
```
prerna-p, line 208: I would probably make it more descriptive like the description for other properties.
```suggestion
            "description": "A list of JSON pointers for properties that are hidden. These properties can be used by developers to test new properties but will not be visible for customers to use",
```

prerna-p (approved):

omkhegde (approved):

ammokhov (approved):

benbridts:
I have some questions:

- How do nonPublicProperties work with contract tests? Are they expected to be always returned? Only sometimes? Is there a test for either behaviour?
- How does that work with Cloud Control Api? Will it be hidden or not hidden there?
- What happens if I use a nonPublicDefinition in a regular property? And vice versa?

WaelA:
> I have some questions:
> 
>     * How do nonPublicProperties work with contract tests? Are they expected to be always returned? Only sometimes? Is there a test for either behaviour?

They are as any other properties. If you specify them in the test input, they will get tested, otherwise they shall not matter.

>     * How does that work with Cloud Control Api? Will it be hidden or not hidden there?

Cloud Control Api will behave same as CloudFormation. The only difference is that nonPublic properties are not documented yet as other properties and shall be used with care as they represent unreleased features.

>     * What happens if I use a nonPublicDefinition in a regular property? And vice versa?

It is all about visibility and what should be available for public to use.

PR #136, opened 2022-04-13
fix a NPE with nested writeOnlyProperties with optional parent.

bcorman3:
*Issue #, if available:*

*Description of changes:*
Change the removeProperty function to catch NullPointerException, as this implies the parent object was null when removing a nested property. This function is intended to take no action when the property to be removed was null, so it should take no action in this case as well.


darrylabbate (approved):

WaelA (approved):

bcorman3:
Updated to use optional so we don't swallow any unexpected exceptions.

darrylabbate (approved):

ammokhov (approved):

PR #146, opened 2023-05-25
Add permissions to tagging schema

yueful27:
Add permissions field in tagging schema to specify permissions needed for tags.

Add unit test for permissions field.

### DIFF
--- a/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
+++ b/src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
@@ -167,7 +167,7 @@ public class ResourceTypeSchema {
                 } else if (key.equals(ResourceTagging.TAG_PROPERTY)) {
                     taggingValue.setTagProperty(new JSONPointer(value.toString()));
                 } else if (key.equals(ResourceTagging.TAG_PERMISSIONS)) {
-                    HashSet<String> tagPermissions = new HashSet<>();
+                    List<String> tagPermissions = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


src/main/java/software/amazon/cloudformation/resource/ResourceTypeSchema.java
```diff
@@ -166,6 +166,10 @@ public ResourceTypeSchema(Schema schema) {
                     taggingValue.setCloudFormationSystemTags(Boolean.parseBoolean(value.toString()));
                 } else if (key.equals(ResourceTagging.TAG_PROPERTY)) {
                     taggingValue.setTagProperty(new JSONPointer(value.toString()));
+                } else if (key.equals(ResourceTagging.TAG_PERMISSIONS)) {
+                    HashSet<String> tagPermissions = new HashSet<>();
+                    ((List<?>) value).forEach(p -> tagPermissions.add(p.toString()));
+                    taggingValue.setTagPermissions(tagPermissions);
```
ammokhov, line 172: do we need to make it `HashSet`? Why dont we use list, then you dont really need to convert it

yueful27: I follow the type definition in `Handler.java` as `Set<String>` and I think the handler permissions is also set as an `Hashset`, but I am not sure whether it is a must for tagging permissions. I can change it to list.

WaelA (approved):

AidenAtSea (approved):

ammokhov (approved):

